PR #37, opened 2017-01-26
Update copyright headers to properly reference initial code.

mickaelistria:


angelozerr:
@mickaelistria there is too classes from https://github.com/angelozerr/textmate.java/tree/master/org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/oniguruma

I don't know if it's OK?

Issue #49, opened 2017-03-07
Drop reference to org.sonatype.oss:oss-parent


mickaelistria:
As the infrastructure is moving to Eclipse.org, maybe there's an opportunity to drop the reference to org.sonatype.oss:oss-parent .
@pascalleclercq Which feature provided by the parent is tm4e using?

pascalleclercq:
@mickaelistria you are perfectly right we have to drop the parent pom.xml I use org.sonatype.oss:oss-parent  for other projects and probably copy/pasted It.
Will do It ASAP.

pascalleclercq:
There's more cleanup that can be done in the pom.xml like the ftp upload configuration. I think we can drop also the pushToBintray.sh.

mickaelistria:
let's keep the pushToBintray.sh until we have builds published fully at Eclipse.org.
We'll add a pushToDownloadEclipseOrg.sh, make it work, and then we can consider dropping bintray.

pascalleclercq:
Ok, sounds good

angelozerr:
@pascalleclercq can we close this issue? It seems it was resolved with https://github.com/eclipse/tm4e/pull/52

pascalleclercq:
Yes, indeed we can close It !

PR #52, opened 2017-03-09
Removed useless reference to org.sonatype.oss:oss-parent

pascalleclercq:
see : https://github.com/eclipse/tm4e/issues/49


angelozerr:
Thansk @pascalleclercq !

PR #53, opened 2017-03-10
Update README with more Eclipse info

mickaelistria:


angelozerr:
Thanks @mickaelistria 

It's out of the scope of this PR, but it seems there is a problem with build with JUnit https://travis-ci.org/eclipse/tm4e/builds/209667947#L1370

PR #62, opened 2017-03-14
Make tests use their specific package

mickaelistria:
To avoid issue regarding signature, if we want to keep using plain
`maven-surefire-plugin` instead of `tycho-surefire-plugin` for the core
parts, we need tests to be in their own package.


angelozerr:
@mickaelistria your PR seems breaking the tests.

mickaelistria:
I did update the PR to avoid the issue.

angelozerr:
Thanks @mickaelistria

PR #64, opened 2017-03-23
Use a target definition

mickaelistria:
a .target file can be used both by Tycho and PDE in the IDE.


angelozerr:
Thanks @mickaelistria !

Issue #63, opened 2017-03-22
bundle as many textmate files as possible to have out of the gate great support


maxandersen:
see $subject

tm4e really should be in EPP together with a library of as many textmate files as possible IMO.

Should not require users to install additional libraries or having to do another tycho plugin buil dsetup just to get started on syntax highlighting imo :)

angelozerr:
Hey @maxandersen ! Glad tm4e could interesest you!

> tm4e really should be in EPP together with a library of as many textmate files as possible IMO.

Totally agree with you, but today the problem is that you can link a TextMate grammar with an editor only with extension point:

 * you define a content type with extension point
 * you link a TextMate grammar with content type with extension point
 * you link TMPresentationReconclier to an editor with Java code or with extension point for Generic Editor.

To fix that, we need to fix GenericEditor issue https://github.com/eclipse/tm4e/issues/17 in other words gives the capability to link TMPresentationReconclier with any content type (`org.eclipse.core.runtime.text`)

Once we have will have this feature user will have the capability to register with TextMate Preferences a grammar  https://github.com/eclipse/tm4e/issues/32 and link to this grammar the `org.eclipse.core.runtime.text` content type.

In other words, I'm waiting for https://bugs.eclipse.org/bugs/show_bug.cgi?id=507753#c2 will be fixed in GenericEditor to work on TextMate Preferences.

We could after provide too a plugin which resgister a lot of TextMate grammar (a catalog kind) binded to the `org.eclipse.core.runtime.text` content type and I think your issue will be fixed, is that?

mickaelistria:
Note that the issue of dynamically (without extension) detecting a textmate grammar for a given document/file isn't a blocker for this one.
We could imagine as a 1st iteration to have tm4e defining all 3 extensions for each of the given grammar. It's indeed a lot to copy-paste, but it's entirely doable already.

mickaelistria:
There are already a few one in the "TM Samples" bundle which is part of the p2 repo. However, those aren't registered in the generic editor.

angelozerr:
@maxandersen just for your information, now you can register your TextMate grammar with user preferences. User must download the TextMate grammar but after you can consume it. Here a little demo:

![textmatepreferencesdemo](https://cloud.githubusercontent.com/assets/1932211/25719317/56a90d7a-3109-11e7-8352-a9be638f4253.gif)

[...]

PR #72, opened 2017-03-27
Issue #63: connect sample grammars to Generic Editor

mickaelistria:

### DIFF
--- a/org.eclipse.tm4e.samples/.classpath
+++ b/org.eclipse.tm4e.samples/.classpath
@@ -6,9 +6,7 @@
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src/main/java/"/>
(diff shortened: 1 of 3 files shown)


angelozerr:
@mickaelistria perhaps we should have a new plugin sample with Generic Editor, no?

mickaelistria:
I don't think it would be that helpful. This plugin is about sample, not production and the dependency to generic editor is optional.
What issue are you trying to avoid by moving the extension in another bundle?

angelozerr:
> What issue are you trying to avoid by moving the extension in another bundle?

If user have no Oxygen, it will have errors in the plugin.xml. More with Generic Editor you need just extension point (no need Java code). It will show quickly that Generic Editor is more extensible than classic TextEditor.

mickaelistria:
@angelozerr if you don't mind, I'd rather see this merged as it, and maybe extracted to a separate bundle later, as this is not very high priority.

angelozerr:
No problem @mickaelistria, thanks!

Issue #69, opened 2017-03-23
Improve Thread sync in AbstractTMModel


mickaelistria:
The AbstractTMModel is a complex class, with a lot of `synchronized`. We should find out ways to simplify it and better separate the thread and the model. I suggest we try to make the `BackgroundThread` static and we avoid to consume fields from AbstractTMModel directly, and try to avoid sharing any reference and pass arguments by value as much as possible. That should progressively remove all needs for `synchronized` and should hopefully improve performance.

angelozerr:
> The AbstractTMModel is a complex class, with a lot of synchronized. 

So ugly class!!! If I remember I have tried to use some pieces of code from `org.eclipse.jface.text.reconciler.AbstractReconciler`  but class is became so uggly -(

> That should progressively remove all needs for synchronized and should hopefully improve performance.

I trust you Mickael!

mickaelistria:
Another issue is that at the moment the `fDirtyRegionsQueue` or `lines` don't track whether there's still job to perform. Instead of checking the `lines`, the thread should check such a queue of regions to process.
This is probably better being tracked in a separate commit.

mickaelistria:
I have the impresion that the _invalidLineStartIndex is enough to check whether to wait or not.

mickaelistria:
I think the invalid lines should be managed by a PriorityBlockingQueue. The model would populate it, and the thread would consume the it. This structure is pretty simple to use and probably the most efficient available to manage a list of lines to process.

angelozerr:
@mickaelistria I trust you, I have none idea about your suggestion.

angelozerr:
@mickaelistria I think we can close this issue, no?

mickaelistria:
Yes, There might be other possible changes, but the current state is somehow what I expected by opening this issue and I do not have an idea of another worthy improvement there.

PR #73, opened 2017-03-27
Issue #69: simplify BackgroundThreadThread

mickaelistria:


mickaelistria:
A (not much reliable) local tests gives:
* without the patch: tsserver.js colorized in 47s
* with the patch: tsserver.js colorized in 32s
However, the test protocol wasn't that safe.
@angelozerr Please check there is at least no slow down on your machine too before merging.

mickaelistria:
Found a major drawback to this patch which is that I see
```
OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x0000000713800000, 165675008, 0) failed; error='Cannot allocate memory' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 165675008 bytes for committing reserved memory.
# An error report file with more information is saved as:
# /home/mistria/hs_err_pid30768.log
```
Much more often.

angelozerr:
@mickaelistria your PR seems great, but please let me time to try it.

Do you know why you have memory problem? Is your PR consumes more memory?

mickaelistria:
I think the memory issue is only for the big tsserver.js file, and that it was also happening previously with a short heap space. I don't think it's a real memory problem as when I enable the heap usage bar in the workbench, I can see the Garbage Collector cleaning up stuff often and after parsing is done, returning to a relatively low value. So I don't think there's a memory leak.
I also believe my proposal doesn't create more objects than the previous one, so it shouldn't impact memory consumption.

angelozerr:
>  I can see the Garbage Collector cleaning up stuff often and after parsing is done

Do you use an eclipse plugin to see that?

> I also believe my proposal doesn't create more objects than the previous one, so it shouldn't impact memory consumption.

Great!

mickaelistria:
> Do you use an eclipse plugin to see that?
No, I just go to Preferences > General and check "Show heap status".

angelozerr:
> No, I just go to Preferences > General and check "Show heap status".

I didn't know that. Very cool feature. Thanks for the info!

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/AbstractTMModel.java
```diff
@@ -128,170 +70,203 @@ public int getSize() {
...
+					model.lines.addLine(line);
+				} catch (Exception e) {
+					e.printStackTrace();
+				}
+			}
+			model.lines.forEach(ModelLine::resetTokenizationState);
+			model.lines.get(0).setState(tokenizer.getInitialState());
+			model._invalidLineStartIndex = 0;
+			
+			while (!interrupted()) {
```
angelozerr: With my Windows OS, even if thread is interrupted (when editor is closed), the while bloc content is executed and my CPU is growing.

mickaelistria: The interrupted state flag is supposed to cover the use-case, there shouldn't have need to add another flag that do that. Locally, on Linux, I can see the Thread running when editor is open (I see it when pausing on the process with the debugger) and the Thread not present any more when closing the editor (same thing, closed editor than paused process to look at threads).
Can you please try with `isInterrupted()` instead of `interrupted()` here?

mickaelistria: And also, are you sure the `dispose()` method was invoked on this TMModel?

angelozerr: Yes I'm sure but I cannot reproduce it -( I will try to redo the problem.

mickaelistria: I'll work on a test case that checks that automatically in the meantime and add it to the review when ready.

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/AbstractTMModel.java
```diff
@@ -128,170 +70,203 @@ public int getSize() {
...
 					}
 				}
-			} while (isDirty);
-		}
-
-		/**
-		 * Reset the background thread as the text viewer has been changed,
-		 */
-		public void reset() {
+				while (model.lines.getSize() >= 0 && model._invalidLineStartIndex < model.lines.getSize()) {
```
angelozerr: When I do open a big file and I do nothing my CPU is growing although I do nothing.

After studying the problem, it seems that when a Java list is big the List#size() can consume a lot CPU.

I have done a simple test:

```
		List<String> list = new ArrayList<>();
		for (int i = 0; i < 999; i++) {
			list.add("eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee");
		}
		
		Thread t = new Thread() {
			@Override
			public void run() {
				while(true) {
					// Uncomment the wait code to avoid CPU problem
					/* synchronized (list) {
						try {
							list.wait(1);
						} catch (InterruptedException e) {
						}
					}*/
					int size = list.size();					
				}
			}
		};
		t.start();
		t.join();
```

With my Windows OS, my CPU is working, but if I uncomment synchronized block which use wait, I have not this problem.

In other words I think we should use wait to avoid computing list size every time and causes CPU problems

mickaelistria: This example you wrote checks size once every second. The proposal of change I wrote makes that an size is checked every 500ms (next instructions is wait for 500ms or notification) when the list is empty. I don't see how the proposal is running into an infinite loop without any wait.

angelozerr: I don't understand the test with lines is empty. You will never this case (even if your editor is empty, line will be initialized with size == 1).

In other word, the following code

`model.lines.wait(500); // release synchronized lock at least every 500ms`

is never executed.

Perhaps you have done a confusion with the AbstractReconciler which fDirtyRegionQueue which store dirty region. In your case lines is not dirty region but the content of the editor.

mickaelistria: ok, got it. The `fDirtyRegionsQueue` should be kept then. I'll restore it.

mickaelistria: Looking at the code before my change, I see `fDirtyRegionQueue=lines` and then nothing changing the value of this queue. Did I miss something? Or does that mean that the bug of continuously running thread was there with previous version of the code?

angelozerr: Here the code which works for me:

```
boolean init = true;
while (!interrupted()) {
	if (!init) {
		synchronized (model.lines) {
			try {
				model.lines.wait(500); // release synchronized lock at least every 500ms
			} catch (InterruptedException x) {
				Thread.currentThread().interrupt();
			}
		}
	}
	init = false;
	while (model.lines.getSize() >= 0 && model._invalidLineStartIndex < model.lines.getSize()) {
		if (interrupted()) {
			return;
		}
		this._revalidateTokensNow(null);
	}
}
```

wait is done every time except when editor is opened.

The call of Thread.currentThread().interrupt(); is very important otherwise the thread is not really interrupted. (It was my problem I think with thread was not interrupted)

angelozerr: Looking at the code before my change, I see `fDirtyRegionQueue=lines`

Again an uggly code.

mickaelistria:
I've made several edits according to your comments. Can you please try again?
About the lack of structure to track whether there's stuff to process, I think it's better being covered in another patch. I'll comment the issue.

mickaelistria:
@angelozerr would adding an `hasInvalidLine()` to ModelLines and checking this as condition to wait or not work?

angelozerr:
It seems that it works like before. Thanks a lot @mickaelistria to clean the code.

angelozerr:
> @angelozerr would adding an hasInvalidLine() to ModelLines and checking this as condition to wait or not work?

To be honnest with you, I don't know really.

PR #110, opened 2017-06-12
Minor improvements

fabioz:
- Instead of creating an UUID, create a simple Object to be compared by
identity.
- Don't add a '\n' if the line already finished with a '\n'.

Signed-off-by: Fabio Zadrozny <fabiofz@gmail.com>

### DIFF
--- a/org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/grammar/Grammar.java
+++ b/org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/grammar/Grammar.java
@@ -12,6 +12,7 @@
  * Contributors:
  *  - Microsoft Corporation: Initial code, written in TypeScript, licensed under MIT license
  *  - Angelo Zerr <angelo.zerr@gmail.com> - translation and adaptation to Java
+ *  - Fabio Zadrozny <fabiofz@gmail.com> - Not adding '\n' on tokenize if it already finished with '\n'
  */
 package org.eclipse.tm4e.core.internal.grammar;
 
@@ -46,7 +47,7 @@ import org.eclipse.tm4e.core.theme.ThemeTrieElementRule;
 
 /**
  * TextMate grammar implementation.
- * 
+ *
  * @see https://github.com/Microsoft/vscode-textmate/blob/master/src/grammar.ts
  *
  */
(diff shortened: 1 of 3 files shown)


org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/grammar/Grammar.java
```diff
@@ -225,8 +225,11 @@ public ITokenizeLineResult2 tokenizeLine2(String lineText, StackElement prevStat
 			isFirstLine = false;
 			prevState.reset();
 		}
-		
-		lineText = lineText + '\n';
+
+		if(lineText.isEmpty() || lineText.charAt(lineText.length()-1) != '\n') {
```
angelozerr: Have you some case where line ends with '\'n'? vscode-textmate does like this.

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/oniguruma/OnigRegExp.java
```diff
@@ -1,23 +1,21 @@
...
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Initial code from https://github.com/atom/node-oniguruma
+ * Initial copyright Copyright (c) 2013 GitHub Inc.
+ * Initial license: MIT
+ *
+ * Contributors:
+ *  - GitHub Inc.: Initial code, written in JavaScript, licensed under MIT license
+ *  - Angelo Zerr <angelo.zerr@gmail.com> - translation and adaptation to Java
```
angelozerr: Please add your name.

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/grammar/Grammar.java
fabioz: Well, right now I'm using the grammar directly, and as I already had a StringBuffer with the contents, I always make sure it ends with a \n -- as it's always required, having clients add it is usually better (to avoid having to allocate a new copy just to add a '\n').

angelozerr:
Thanks @fabioz !

Issue #17, opened 2016-11-23
Defines "org.eclipse.core.runtime.text" GenericEditor


angelozerr:
Once GenericEditor presentationReconcilers will support "org.eclipse.core.runtime.text" to bind IPresentationReconclier (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=507772), the org.eclipse.textmate4e.ui could declare in the plugin.xml:

```
<extension
         point="org.eclipse.ui.genericeditor.presentationReconcilers">
      <presentationReconciler           class="org.eclipse.language.textmate.eclipse.text.TMPresentationReconciler"
            contentType="org.eclipse.core.runtime.text">
```

It will give the capability to register a TextMate grammar with UI preferences and when user will open a file with GenericEditor, the syntax coloration should be done.

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=507753#c2

mickaelistria:
I've proposed a patch for https://bugs.eclipse.org/bugs/show_bug.cgi?id=507772 . With this, you can simply have the TMPresentationReconciler throw an IllegalArgumentError in the `install(...)` method (according to the provided viewer and document) to notify the generic editor that it should be ignored.
It would be great if you could give it a try.

mickaelistria:
In order to have this, TMPresentationReconciler mustn't show the pop-up message in case of missing grammar. I believe this piece of logic, showing the pop-up in case of missing grammar, makes more sense inside the TMEditor or someplace else that doesn't cascade to generic editor.
Also, TMPresentationReconciler shouldn't throw an error when there is no grammar. In case there is no available grammar for the input, it should simply do nothing. Currently, we cannot associate to text as it would make the Generic Editor show an error and be unusable 
I'm working on another appraoch that would allow making the TMPresentationReconciler the default (if no better one is defined). It's simpler and better, but it requires TMPresentationReconciler to not break things.

angelozerr:
> Also, TMPresentationReconciler shouldn't throw an error when there is no grammar.

When you register with extension point a grammar (with bad information), I think it's cool to see the error in the editor, no?

mickaelistria:
Yes, in case the grammar is erroneous, it's fine to show an error because either it happens during devleopment time and the developer can fix it, or it's a user-defined association and user can remove it.

mickaelistria:
[...]

PR #129, opened 2017-08-07
Issue #17 - Associate with text content-type

mickaelistria:
And throw an exception in install if it cannot apply properly (missing
grammar).

### DIFF
--- a/org.eclipse.tm4e.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.tm4e.ui/META-INF/MANIFEST.MF
@@ -15,7 +15,8 @@ Require-Bundle: org.eclipse.tm4e.core,
  org.eclipse.ui.ide;resolution:=optional,
  com.google.gson,
  org.eclipse.e4.ui.css.swt.theme,
- org.eclipse.core.expressions
+ org.eclipse.core.expressions,
+ org.eclipse.ui.genericeditor;bundle-version="1.1.0";resolution:=optional
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Export-Package: org.eclipse.tm4e.ui,
  org.eclipse.tm4e.ui.internal.model;x-friends:="org.eclipse.tm4e.ui.tests",
--- a/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/text/TMPresentationReconciler.java
+++ b/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/text/TMPresentationReconciler.java
@@ -16,7 +16,9 @@ import java.util.List;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.content.IContentType;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.IPreferenceChangeListener;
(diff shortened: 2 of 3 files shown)


org.eclipse.tm4e.ui/META-INF/MANIFEST.MF
```diff
@@ -15,7 +15,8 @@ Require-Bundle: org.eclipse.tm4e.core,
  org.eclipse.ui.ide;resolution:=optional,
  com.google.gson,
  org.eclipse.e4.ui.css.swt.theme,
- org.eclipse.core.expressions
+ org.eclipse.core.expressions,
+ org.eclipse.ui.genericeditor;bundle-version="1.1.0";resolution:=optional
```
angelozerr: You don't need this dependency, no?

mickaelistria: Not really. However, I believe if we include that patch as it, it would fail with generic editor 1.0.0. That's why I preferred to explicit the version of the dependency here, but still making it optional so it's not blocking on earlier versions.

angelozerr: That's strange, I though it worked without this dependency. I will retry it.

mickaelistria: Oh yeah, I added the `viewer.getDocument() != null` condition to prevent it from failing with older version of generic editor.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/text/TMPresentationReconciler.java
```diff
@@ -420,7 +422,21 @@ private void themeChange(ITokenProvider oldTheme, ITokenProvider newTheme, IDocu
...
+			try {
+				ContentTypeInfo info = ContentTypeHelper.findContentTypes(viewer.getDocument());
+				if (info != null) {
+					this.grammar = findGrammar(info);
+				}
+			} catch (CoreException e) {
+				TMUIPlugin.getDefault().getLog().log(new Status(IStatus.ERROR, TMUIPlugin.PLUGIN_ID, e.getMessage(), e));
+			}
+			if (this.grammar == null) {
+				throw new IllegalArgumentException("No TextMate Grammar for provided editor/viewer/document.");
```
angelozerr: My fear is that with current Eclipse Oxygen, Generic Editor will not work if you have not configured the grammar, isn't it?

If it that perhaps we should not throw this error if editor is a Generic Editor with version < M4 (when your patch will be accepted).

mickaelistria: With Oxygen, `viewer.getDocument() != null` always return `false`, so this block is never executed and the exception never thrown

mickaelistria:
I'm not sure about this PR, please don't merge it!

angelozerr:
> I'm not sure about this PR, please don't merge it!

Yes, I know:) There is a problem with dialog which is opened when grammar is not found. It's very helpful when user wishes open a file with TextMate Editor but not with Generic Editor. We must disable this dialog for Generic Editor or better provide a preference to know if dialog must be opened or not when grammar is not found.

angelozerr:
@mickaelistria I have enable the dialog grammar import wizard only with TextMateEditor. So now with GenericEditor there is not this dialog. 

Could you restest it please.

angelozerr:
@mickaelistria I close your PR in favor of https://github.com/eclipse/tm4e/issues/17

Now Generic Editor is configured with TMPresentationReconciler by default. None error are thrown now.

PR #142, opened 2017-09-09
Minor change in consumer names

mickaelistria:
According to Eclipse.org trademark `Eclipse XXX` is reserved for Eclipse.org projects only. BlueSky isn't (yet) an Eclipse.org project, whereas aCute is.


angelozerr:
Thank's @mickaelistria !

Issue #134, opened 2017-09-04
Code marked as comment due to Umlauts


NielsNet:
There seems to bee an issue with Textmate and umlauts.  
As shown in the issue angelozerr/typescript.java#203 as soon as two or more letters with an umlaut are used the following lines of code are marked as comment even though they should not.  
The following screenshot shows different cases where the behaviour appears.
![grafik](https://user-images.githubusercontent.com/30441612/29863247-5a355f4e-8d6f-11e7-9879-abd29af604ca.png)

angelozerr:
I'm not sure, but perhaps problem comes from that tm4e doesn't support onigurama with UTF-16 (see ignored tests at https://github.com/eclipse/tm4e/blob/master/org.eclipse.tm4e.core.tests/src/main/java/org/eclipse/tm4e/core/grammar/GrammarSuiteTest.java#L44)

I think we should try to implement ConvertUtf16OffsetToUtf8 https://github.com/atom/node-oniguruma/blob/master/src/onig-searcher.cc#L4 in Java at https://github.com/eclipse/tm4e/blob/master/org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/oniguruma/OnigSearcher.java

Any PR are welcome!

fabioz:
I already have some utilities for converting to/from bytes/multibyte I used in LiClipseText... (https://github.com/fabioz/LiClipseText/blob/master/plugins/org.brainwy.liclipsetext.editor/src/org/brainwy/liclipsetext/editor/partitioning/Utf8WithCharLen.java).

I'll take a look on porting them to `OnigString` (but will leave it up to you to where to place the calls, because it's not just convert one way, the result has to be converted back too...).

Maybe better would be just using always utf-8 internally and convert just the final result? Anyway, will provide a pull request for the conversions shortly, but will leave it up to you where to do the `convertUtf16OffsetToUtf8` and `convertUtf8OffsetToUtf16` calls.

angelozerr:
@NielsNet could you reinstall http://download.eclipse.org/tm4e/snapshots/ and tell me if it fixes your problem. Thanks!

fabioz:
@angelozerr With this change I have the same behavior I had with the approach I was using which just used everything with utf-8 offsets internally and just converting the final result from utf-16 to utf-8.

[...]

PR #144, opened 2017-09-15
Fix recursion error on OnigRegExp with unicode chars

fabioz:
Fix for: https://www.brainwy.com/tracker/LiClipseText/3


angelozerr:
@fabioz thanks for your PR. AT first I think you must sign the PR to avoid having problem with IP validation.

I tell me if a better fix could be handle by https://github.com/eclipse/tm4e/issues/134#issuecomment-327043827 like node-onigurama does?

fabioz:
@angelozerr I ended up creating a new pull request (https://github.com/eclipse/tm4e/pull/145) as I had done this one editing through the github interface, so, properly created a branch on my fork to fix it with proper user/email/signature.

Related to the issue you posted, it's a separate issue... the error I had was that it wasn't even able to load a grammar because it had a unicode quote (https://github.com/textmate/cmake.tmbundle/blob/master/Syntaxes/CMake%20Listfile.tmLanguage#L66), not on interpreting the grammar later on.

PR #145, opened 2017-09-16
Fix recursion error on OnigRegExp with unicode chars

fabioz:
Fix for: https://www.brainwy.com/tracker/LiClipseText/3


angelozerr:
Thanks @fabioz !

Issue #136, opened 2017-09-05
Problem with HTML coloring


mniewrzal:
I found HTML file where syntax coloring doesn't work to good. I attached screenshot and example file. Example works fine in VS Code and no stacktraces in logs.

![screenshot from 2017-09-05 10-10-56](https://user-images.githubusercontent.com/7313031/30051585-e25f6f6a-9222-11e7-8513-0e5f67e76ef6.png)

[test.txt](https://github.com/eclipse/tm4e/files/1276555/test.txt)

angelozerr:
I suspect it's the same issue than https://github.com/eclipse/tm4e/issues/134#issuecomment-327043827 (encoding problem)

angelozerr:
@mniewrzal could retry with last version of tm4e, thanks!

angelozerr:
@mniewrzal  can we close this issue?

mniewrzal:
Yes, look ok now :) Sorry for the delay, I was far from Eclipse things for some time.

:+1:

PR #146, opened 2017-09-16
Utilities to convert to/from utf-8, utf-16.

fabioz:
First step to solve https://github.com/eclipse/tm4e/issues/134

### DIFF
--- a/org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/oniguruma/OnigString.java
+++ b/org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/oniguruma/OnigString.java
@@ -13,11 +13,15 @@
  *  - GitHub Inc.: Initial code, written in JavaScript, licensed under MIT license
  *  - Angelo Zerr <angelo.zerr@gmail.com> - translation and adaptation to Java
  *  - Fabio Zadrozny <fabiofz@gmail.com> - Convert uniqueId to Object (for identity compare)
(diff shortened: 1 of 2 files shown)


fabioz:
As a note, with these differences, I've done a change to just convert indexes back to utf-16 and things almost seem to work.

You can see my changes at:

https://github.com/fabioz/LiClipseText/commit/1d8f88c82676d30edf35b1ef8a70803df3f2681e 

So, mostly I just convert the result back to utf-16 (as everything should be dealing with utf-8 internally), now, I have a strange difference in that it appears that the grammar found an intermediary match that it didn't find first, so, I'm not submitting a pull request because of that...

i.e.: 
I expected my test to have the same results with chars with and without unicode chars, yet, as you can see in: 
https://github.com/fabioz/LiClipseText/commit/1d8f88c82676d30edf35b1ef8a70803df3f2681e#diff-9b217bdbf9f5aee0083ab4831116a634R62

it's not really exactly the same (the assertEquals above is the one without unicode and the below is with unicode -- the result should be the same).

angelozerr:
Thanks a lot @fabioz for your PR. Let me find time to study it and compare with VSCode.

angelozerr:
@fabioz I hope I have fixed problem with UTF-16 (see https://github.com/eclipse/tm4e/issues/134) 

I have tested https://github.com/eclipse/tm4e/issues/134 and https://github.com/eclipse/tm4e/issues/136 and it seems it works well in Windows context. Please give me feedback.

I'm using your convert methods which works well, Thanks!

PR #171, opened 2018-08-23
Language Configurations Errors

LucasBullen:
- Discovered multiple errors while testing with Corrosion
 - Misunderstood on enter actions, now an extra indent is only added
when the insert text is empty
 - De-toggling a group of line comments would cause errors when one of
the lines were empty
 - Now accounts for multi-letter auto closing pairs (/**, */)


org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/ToggleLineCommentHandle.java
```diff
@@ -184,7 +184,7 @@ private void removeLineComments(IDocument document, ITextSelection selection, St
 		Boolean isStartBeforeComment = false;
```
angelozerr: Please rename class to ToggleLineCommentHandler (with missing 'r')

Issue #172, opened 2018-08-23
Comment support doesn't work on Windows OS + CSS Language Server


angelozerr:
@LucasBullen I create this issue, because your work on comment support doesn't work on a given context : 

 * Windows OS where Ctrl+Shift+/ doesn't work (in Windows OS I must key on Shift to have '/'). I have the same problem with JDT. Because of this problem, I can never call block comment which is only supported by CSS and XML, HTML language.
 * CSS Language Server which defines only block comment.

1) Key command problem 

So to fix the key command problem, I suggest to do like WTP:

 * for adding/removing block comment, WTP does like this:

```xml
<key
	sequence="M1+M2+/"
	contextId="org.eclipse.wst.sse.ui.structuredTextEditorScope"
	commandId="org.eclipse.wst.sse.ui.add.block.comment"
	schemeId="org.eclipse.ui.defaultAcceleratorConfiguration">
</key>
<key
	sequence="M1+/"
	contextId="org.eclipse.wst.sse.ui.structuredTextEditorScope"
	commandId="org.eclipse.wst.sse.ui.add.block.comment"
	schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
	platform="carbon">
</key>
<key
	sequence="M1+M2+\"
	contextId="org.eclipse.wst.sse.ui.structuredTextEditorScope"
	commandId="org.eclipse.wst.sse.ui.remove.block.comment"
	schemeId="org.eclipse.ui.defaultAcceleratorConfiguration">
</key>
<key
	sequence="M1+\"
	contextId="org.eclipse.wst.sse.ui.structuredTextEditorScope"
	commandId="org.eclipse.wst.sse.ui.remove.block.comment"
	schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
	platform="carbon">
</key>
```
 * for adding/removing line comment, WTP does like this:

```xml
<key
	sequence="M1+M2+C"
	contextId="org.eclipse.wst.sse.ui.structuredTextEditorScope"
	commandId="org.eclipse.wst.sse.ui.toggle.comment"
	schemeId="org.eclipse.ui.defaultAcceleratorConfiguration">
</key>
```
Are you OK @LucasBullen, @mickaelistria to do that? It will work on any OS.

2)  CSS Language Server which defines only block comment.

[...]

PR #173, opened 2018-08-23
Issue #172 Comment support doesn't work on Windows

LucasBullen:
- Use generic key squences to work across systems


angelozerr:
Thanks @LucasBullen ! It works better!

Issue #174, opened 2018-08-23
Contextual menu for comments commands


angelozerr:
It should be cool to have comments commands in contextual menu (like WTP or JDT)

![image](https://user-images.githubusercontent.com/1932211/44549826-9a0bee80-a722-11e8-9c70-86c914aa5d8e.png)

We can see in this case the key binding to use in the menu.

@LucasBullen is there any chance that you work on this issue?

mickaelistria:
The source menu is contributed by JDT or WTP. Should we first add a "Source" menu definition to the Generic Editor?

angelozerr:
> Should we first add a "Source" menu definition to the Generic Editor?

Perhaps yes?

sebthom:
Implemented via #421

PR #175, opened 2018-08-24
Issue #174 Contextual menu for comments commands

LucasBullen:
- Relies on https://git.eclipse.org/r/#/c/128034/
 - Creates popup option for the comment commands for users to quickly
see the key activation sequences of each command
 - Fixes small issue with mulit-char auto inserts moving the caret too
far


LucasBullen:
@angelozerr https://git.eclipse.org/r/#/c/128034/ is not yet merged into master so this patch won't currently work, won't cause any visual errors though.

angelozerr:
Ok thanks for the hint. I will try when it will be available.

PR #182, opened 2018-10-11
Bug 540050 - NPE in HasLanguageConfigurationPropertyTester

briandealwis:
ContentTypeHelper.findContentTypes(document) may return null.


mickaelistria:
Obviously a good patch, thanks!

Issue #163, opened 2018-07-06
Set dark theme to all grammars


PhaserEditor2D:
Hi,

Since Photon release Eclipse can use the Linux theme, so now users can have a dark Eclipse without selecting the dark Eclipse theme. But TM4E does not have a way to know if the OS theme is dark or not, so it shows by default the light theme associate to the grammars.

Right now I am using Adwaita dark Gnome theme, and I have to set manually a dark theme to all the TM4E grammars. It would be great if we have an option to set the default theme for all the grammars. Or maybe a better solution could be to set TM4E in dark mode without set Eclipse in dark mode.

angelozerr:
Eclipse IDE theme doesn't provide some category like this thme is a dark kind them. That's why I have hard coded that in https://github.com/eclipse/tm4e/blob/master/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/AbstractThemeManager.java#L103

I think you should hard coded that to support your dark theme. Any PR are welcome!

sebthom:
This should be solved with 0.10.0. Feel free to reopen if not.

Issue #176, opened 2018-08-26
TM4E overrides manually defined editor colors


mickaelistria:
I've set a background color to my text editors to something that's a bit softer than plain white. I've done it through Preferences > General > Editors > Text Editors changing the background color.
When I open the Generic Editor without a TM4E contributing, all is fine, I get the background color I configured. Whenever I open it on a file that has a TextMate grammar associated and TM4E enabled for it, the color is restored to plain white, reducing my comfort and making this behavior feel a bit buggy.

In case user overrides the setting, it should most likely be honored by TM4E. Same goes with the default foreground color.

jabby:
IIUC, the priority for colors in text editors should be 

1. User definition in Text Editor
2. Textmate grammar if there is no user definition
3. Eclipse Theme if there is nothing else

I didn't look into TM4E code base but I think we should have a reconciler / aggregator for all this definition of color.

mickaelistria:
I could verify that at least the background color is kept and was happy with the result. Enough to close this ticket IMO.

PR #184, opened 2018-10-24
Issue #163: Support another dark themes

Aloomenon:
Hello! 
I add support for devStyle and OS-dependent dark themes, so now themes correctly switch from light to dark and back.

### DIFF
--- a/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/AbstractThemeManager.java
+++ b/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/AbstractThemeManager.java
@@ -17,10 +17,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import org.eclipse.core.runtime.preferences.IEclipsePreferences;
-import org.eclipse.core.runtime.preferences.InstanceScope;
+import org.eclipse.tm4e.ui.TMUIPlugin;
 import org.eclipse.tm4e.ui.themes.ITheme;
 import org.eclipse.tm4e.ui.themes.IThemeAssociation;
+import org.eclipse.tm4e.ui.themes.IThemeBehaviour;
 import org.eclipse.tm4e.ui.themes.IThemeManager;
 import org.eclipse.tm4e.ui.themes.ThemeAssociation;
 
@@ -30,19 +30,14 @@ import org.eclipse.tm4e.ui.themes.ThemeAssociation;
  */
 public abstract class AbstractThemeManager implements IThemeManager {
 
-	// Theme for E4 CSS Engine
-	public static final String E4_THEME_ID = "themeid"; //$NON-NLS-1$
-
-	private static final String E4_CSS_THEME_PREFERENCE_ID = "org.eclipse.e4.ui.css.swt.theme"; //$NON-NLS-1$
-	private static final String E4_DARK = "org.eclipse.e4.ui.css.theme.e4_dark";
-	private static final String DEVSTYLE_DARK = "com.genuitec.eclipse.themes.dark";
-
+	private final IThemeBehaviour behaviour;
 	private final Map<String /* theme id */ , ITheme> themes;
 	private final ThemeAssociationRegistry themeAssociationRegistry;
 
 	public AbstractThemeManager() {
 		this.themes = new LinkedHashMap<>();
 		this.themeAssociationRegistry = new ThemeAssociationRegistry();
+		this.behaviour = TMUIPlugin.getThemeBehaviourManager().getThemeBehaviour();
 	}
 
 	@Override
@@ -68,7 +63,7 @@ public abstract class AbstractThemeManager implements IThemeManager {
 
 	@Override
 	public ITheme getDefaultTheme() {
-		boolean dark = isDarkEclipseTheme();
+		boolean dark = behaviour.isDarkEclipseTheme();
 		return getDefaultTheme(dark);
 	}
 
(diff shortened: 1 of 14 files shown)


angelozerr:
I would like to avoid having hard coding non standard eclipse dark theme. We need I think an extension point or othet thing (preferences) to define a theme as dark.

mickaelistria:
I agree with Angelo aboid avoiding references to specific non Eclipse themes.
The best thing for theming in general is to avoid hardcoding anything or assuming anything about the theme and directly look at the colors of existing and parent widgets to make assumptions and derive colors. Like suggested in #176.

Aloomenon:
I decided to add an extension point that allows you to influence the default theme applying behavior

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/DefaultThemeBehaviour.java
```diff
@@ -0,0 +1,20 @@
+package org.eclipse.tm4e.ui.internal.themes;
```
angelozerr: Why not using your extension point to register DefaultThemeBehaviour ?

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/ThemeBehaviourManager.java
```diff
@@ -0,0 +1,62 @@
...
+	 * @param element
+	 *            configurationElement for extension point
+	 * @return behaviour from configuration or null if exception occured
+	 */
+	private IThemeBehaviour getBehaviour(IConfigurationElement element) {
+		IThemeBehaviour extendedBehaviour = null;
+		String className = element.getAttribute("class");
+		String bundleName = element.getContributor().getName();
+		try {
+			Class<?> clazz = Platform.getBundle(bundleName).loadClass(className);
```
angelozerr: Why not using element.createExecutableExtension("class") ?

angelozerr:
@Aloomenon thanks for your PR, I like it. I have added 2 comments.

@mickaelistria what do you think about this new extension point? Please review it if you have find time. Thanks!

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/DefaultThemeBehaviour.java
Aloomenon: I'd prefer to always have some default behaviour. I think that if we will register it, ThemeBehaviourManager.init() should change to 
`while (behaviour == null && configurationElements.length > 0) {
behaviour = getBehaviour(configurationElements[i]);
i++;
}` or something like that, where we occure exception and we wont have any behaviour.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/ThemeBehaviourManager.java
Aloomenon: yep, that's more correct. Changed in [8abb303](https://github.com/eclipse/tm4e/pull/184/commits/8abb303343e5ecd565bb52aca7d9efc4afdfdb16) commit.

mickaelistria:
What is a ThemeBehavior? I'm not convinced the wording is explicit enough here.
What's the purpose of this, what is it actually doing? Is it providing a specific TM theme that is enabled according to the current Eclipse IDE theme? 

I also think we should really avoid hardcoding behavior related to a theme or another. TM4E is not the right place where to store DevStyle-specific theme configuration for instance. On that regard, an extension-point makes sense.

I think this also relates to bugs like #176. We should probably fix those bugs before designing new extensions. With bug fixed, it could reduce the scope of the extensions and make them more maintainable.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/IThemeBehaviour.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * API to identify dark or light color scheme applied.
+ * 
+ * Motivation: get the opportunity to influence the applying TM themes behavior depending on the eclipse theme.
+ * 
+ * @author Tim Kovalyov
+ *
+ */
+public interface IThemeBehaviour {
```
mickaelistria: The world of themes is not so binary, it's not dark vs light. It can be blue, yellow, grey... I'm not convinced an API which as such a binary interpretation of theming will scale with time and use-cases.
Why instead of invoking an API to retrieve the theme, TM4E should look at the default background color of the widget and decide whether it's dark or not? Such more dynamic behaviour could prevent from introducing an API just to know whether current theme is dark or not.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/AbstractThemeManager.java
```diff
@@ -1,157 +1,134 @@
...
-	}
-
-	@Override
-	public boolean isDarkEclipseTheme() {
-		return isDarkEclipseTheme(getPreferenceE4CSSThemeId());
-	}
-
-	@Override
-	public boolean isDarkEclipseTheme(String eclipseThemeId) {
-		return E4_DARK.equals(eclipseThemeId);
```
mickaelistria: Instead of all this change, to fix the initial request, we can just change it to `eclipseThemeId != null && eclipseThemeId.toLowerCase().contains("dark")`.

Aloomenon: I think we can close that PR because it's already in #176

Aloomenon:
since eca28bf this PR isn't relevant

PR #188, opened 2018-11-08
Issue #176 Add priority to user-defined colors

Aloomenon:


angelozerr:
Thanks @Aloomenon ! I like your PR.

@mickaelistria, what do you think?

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/ColorManager.java
```diff
@@ -47,4 +49,72 @@ public void dispose() {
...
+
+	/**
+	 * The method verifies that a color is defined in a preferences store using a token.
+	 * 
+	 * @param tokenId
+	 *            name of the token
+	 * @return color is user defined or not
+	 */
+	public boolean isColorUserDefined(String tokenId) {
+		return !PreferenceUtils.getEditorsPreferenceStore().getBoolean(getSystemDefaultToken(tokenId), true);
```
mickaelistria: Could we implement this without introducing a "SytemDefault"-based approach? Something like assuming that if a token is not part of the preferences, it means it's system default?

mickaelistria: Also, shouldn't be stored in another prefernece store than the default Editor one? As only TM4E uses them, it would better have them in TM4E preference store IMO.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/Theme.java
```diff
@@ -136,23 +150,9 @@ public boolean isDefault() {
...
-		Color selectionBackground = tokenProvider.getEditorSelectionBackground();
-		if (selectionBackground != null) {
-			styledText.setSelectionBackground(selectionBackground);
-		}
-		Color selectionForeground = tokenProvider.getEditorSelectionForeground();
-		styledText.setSelectionForeground(null);
-		if (selectionForeground != null) {
-			styledText.setSelectionForeground(selectionForeground);
-		}
+		styledText.setBackground(getEditorBackground());
```
mickaelistria: Couldn't `getEditorBackground()` return null if there is no need to override it?
I don't think TM4E should call setBackground at all if it has no good reason to do it (and same for other colors)

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/utils/PreferenceUtils.java
```diff
@@ -0,0 +1,42 @@
+package org.eclipse.tm4e.ui.utils;
+
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.InstanceScope;
+import org.eclipse.tm4e.ui.internal.preferences.PreferenceConstants;
+
+public class PreferenceUtils {
```
mickaelistria: We should probably have a `getTM4EPreferenceStore()`

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/ColorManager.java
Aloomenon: Do you suggest to store default values for all 5 variables used here and check them when it needs?

mickaelistria: Ok, I see how SytemDefault is used in other places of Platform and your usage is consistent with it.
You can ignore my previous comments.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/Theme.java
Aloomenon: hm, getters can return null. I see it in that way :  userDefined >> themeColor >> null. If you have another vision, please, tell me.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/utils/PreferenceUtils.java
Aloomenon: Ok, I'll add it.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/Theme.java
mickaelistria: Yes, but if StyledTest already has defined color that come from someplace else, and that userDefined == null && themeColor == null, then the default color pre-set in the StyledTest should be kept. Here it would remove it.

Aloomenon: Ok, tomorrow I'll push all changes.

Aloomenon:
I've pushed changes. @mickaelistria can u check it ?

Aloomenon:
Additionally, I change getEditorCurrentLineHighlight() logic for this way: if background was user-defined -> take color from prefs, otherwise take it from TM theme

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/Theme.java
```diff
@@ -1,158 +1,179 @@
...
+		this.name = name;
+		this.dark = dark;
+		this.isDefault = isDefault;
+	}
+
+	public Theme(IConfigurationElement ce) {
+		super(ce);
+		id = ce.getAttribute(XMLConstants.ID_ATTR);
+		name = ce.getAttribute(XMLConstants.NAME_ATTR);
+		dark = "true".equals(ce.getAttribute(DARK_ATTR));
```
mickaelistria: Boolean.parse(...)

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/Theme.java
```diff
@@ -1,158 +1,179 @@
...
+		return tokenProvider;
+	}
+
+	@Override
+	public String toCSSStyleSheet() {
+		return super.getResourceContent();
+	}
+
+	@Override
+	public boolean isDark() {
```
mickaelistria: Would be good to find something.more dynamic, like some.`match(themeId)` or `match(Color backgroundColor)` to allow more cases.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/Theme.java
Aloomenon: I pushed last commit with crlf enabled :<
Add correct paring in 360e67f commit.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/themes/Theme.java
Aloomenon: I agree with you, but we should change it later, not in this PR.

Aloomenon:
Edited my previous commit ab31163 by change crlf to lf back. Changes in 6c36e0f

mickaelistria:
This change set looks good to me.
We need cleaner IP processes checks.
@Aloomenon Please squash the commits into a single one as add the "Signed-Off-By" with your committer email. You also need to sign the Eclipse Contributor Agreement as explained in https://dev.eclipse.org/eclipse-webhook/services/status_details.php?id=5be7e8238f8b6 . We're basically not legally allowed to merge it before you do so.

mickaelistria:
@Aloomenon What would be great is if you can also add an automated test for it (at least one of the colors) in org.eclipse.tm4e.ui.tests. Just a test changing the background preference, opening an editor, retrieving the widget (`(StyledText)editor.adapt(Control.class)`) and checking the color seems to be easy enough to implement and would provide a very good coverage of your change.

Aloomenon:
Squashed to 1 commit, ip-validation is ok. 
I'll write a few tests next week and push them asap

Aloomenon:
@mickaelistria, @angelozerr 
All is ready. Tests in commit.

mickaelistria:
Looks good to me.
@angelozerr what do you think?

Issue #192, opened 2018-12-03
Typing closing character in an open/pair should just move caret if next char is the typed one


mickaelistria:
The implementation of closing bracket or quote sometimes inserts more than desired when typing quickly.

I suggest we refine the strategy so that if we are in a matching pair, we will ignore (auto-delete) the insertion of any redundant character. For instance ( `|` is the caret)
* `"|"` and I type `"` then the auto-edit would notice that the typed `"` is probably the closer of the `""` pair (since previous char is opener) and would see than next char after caret already closes the pair -> ignore typed character
* `""|` and I type `"`->  char[offset-2] and char[offset - 1] form a pair and inserted char is supposed to close the pair -> ignore typed character.

mickaelistria:
@angelozerr it would be great to have it for 4.10 (ie this week). Do you think it's achievable?

angelozerr:
@mickaelistria I have similar behvaiour in my workspace that I had not pushed. I pushed it now. The behaviour is not exactly your issue, but it does the same thing than vscode. Here a little demo:

![autoclose](https://user-images.githubusercontent.com/1932211/49364477-b4bc0f00-f6e4-11e8-830d-8a8fc081f01e.gif)

Hope you will like it.

mickaelistria:
This looks good. @norru @ctron what do you think?

ctron:
So, I just tried this with "TextMate Core	0.3.0.201812030915, tm-feature.feature.group. Angelo ZERR":

It is definitely an improvement. However, it only works for the first level of auto-closing.

Assume I type in `foo("bar")` … then the autoclosing works fine for the second `"`, but fails for the closing `)`.

![gh-issue](https://user-images.githubusercontent.com/202474/49375911-e6919d80-f705-11e8-9cc4-69bc7de30b71.gif)

mickaelistria:
We can maybe make things simpler: whenever there is a closing character after the caret, we do not proceed the insertion of this character and only move the caret. It doesn't seem to me this would cause any big trouble at usage.

mickaelistria:
This was improved with #196 and #194 . I'm closing it, please open new tickets for other cases (ideally 1 ticket for each case) where auto-edit is annoying.

PR #196, opened 2019-01-08
Issue #192 - Ignore insertion of repeated closing char

mickaelistria:
In order to simplify the flow when a user is likely to type a character
that was auto-inserted earlier, like a ) after user typed ( while user
was typing (), we just ignore typed auto-close characters if it's
already in the document and we only move the caret.


mickaelistria:
@theDragonFire Can you please have a look and check whether this covers your use case (at least the test matches your expectation)?

theDragonFire:
Yep, that is what I was looking for. Thanks!

norru:
Severe regression.

If autoclose is disabled, typing "{}" produces only "{".

With "Autoclose" enabled, typing "()))))))" produces "()".

The workaround is to type the affected line *in another editor* and copy-paste it. to Eclipse.

norru:
Should I create a new ticket?

Issue #160, opened 2018-06-25
Highlight TableViewer cells with TM4E


PhaserEditor2D:
Hi,

I need to create a tm4e JavaScript tokenizer to render the content of a TableViewer cell.

My tool uses Eclipse BlueSky (from @mickaelistria ) and the idea is to get the current TM theme associated with the JavaScript grammar, create a tokenizer and transform the result in an array of JFace StyleRange, then I can "syntax paint" the cell.

Any tip is welcome!

I attach here an image of the table:

![image](https://user-images.githubusercontent.com/12187299/41865933-af0be81e-787c-11e8-95c9-cdf185fb611e.png)

mickaelistria:
This is a very interesting and not so easy question ;)

Have you seen some occurences of other languages (like Java with JDT) which managed to reuse the syntax highlighting in other widgets than StyledText? If so, it could simply be a matter of mimicking their implementation, using the TMPresentationReconciler instead.
But to be honest, I'm not sure anyone ever did that. If I see something of that kind, I'll add a note here.

Another possibility is simply to find the right code from tm4e to add into the LabelProvider. It might be simple enough, but there are some risks of this taking too much time in a synchronous UI Thread operation, then causing freezes in the IDE.

PhaserEditor2D:
I just did it! :)

![image](https://user-images.githubusercontent.com/12187299/41874394-d44065ec-7895-11e8-8bb4-8bab37756e44.png)

I can go home now. Tomorrow I will work more on this. This is the important code:

```

private StyleRange[] parseCode(String line) {
	List<StyleRange> ranges = new ArrayList<>();

	ITokenizeLineResult result = _grammar.tokenizeLine(line);
	for (IToken token : result.getTokens()) {
		for (String scope : token.getScopes()) {
			org.eclipse.jface.text.rules.IToken themeToken = _theme.getToken(scope);
			if (themeToken != null) {
				Object data = themeToken.getData();
				if (data != null && data instanceof TextAttribute) {
					TextAttribute attr = (TextAttribute) data;
					if (attr.getForeground() != null || attr.getBackground() != null) {
						StyleRange range = new StyleRange(token.getStartIndex(),
								token.getEndIndex() - token.getStartIndex(), attr.getForeground(),
								attr.getBackground());
						ranges.add(range);
					}
				}
			}
		}
	}

	return ranges.toArray(new StyleRange[ranges.size()]);
}


```

PhaserEditor2D:
[...]

PR #198, opened 2019-01-12
Add Phaser Editor to the list of tools

PhaserEditor2D:
Add Phaser Editor to the list of tools that uses tm4e.

Signed-off-by: PhaserEditor2D <developers@phasereditor2d.com>


PhaserEditor2D:
@mickaelistria I get it from WWD but I use it in other tasks out of the scope of WWD (I guess), like paint matching code lines in a custom docs search tool we have, named Chains:

image

You can see here in this code we use only TM4E API: https://github.com/PhaserEditor2D/PhaserEditor/blob/7bcfce9d47f5483e215cfa8e783e7a61588c6397/source/v2/phasereditor/phasereditor.chains.ui/src/phasereditor/chains/ui/views/ChainsView.java#L593

PhaserEditor2D:
This is an old issue about it:

https://github.com/eclipse/tm4e/issues/160

PR #215, opened 2019-07-22
Align Eclipse-light theme with the classic Eclipse colors

jonas-jonas:
This PR aligns the colors of the Eclipse-light theme with the colors of the classic Eclipse Java editor theme. I tested it locally with a Dart file. See images for comparisons. 

*Without* my patch: 
eclipse-light-old

*With* my patch (see "void" keyword and less yellow, which is not part of the origin color theme):
eclipse-light-new

Classic Eclipse Java colors: 
eclipse-light-original

Signed-off-by: Jonas Hungershausen <jonas.hungershausen@vogella.com>


mickaelistria:
Very good, thank you!

Issue #149, opened 2017-12-05
Cannot remove previously added grammar


creckord:
To toy around with TM4E, I downloaded [XML.plist](https://raw.githubusercontent.com/textmate/xml.tmbundle/master/Syntaxes/XML.plist) into my workspace and registered it through `Preferences / TextMate / Grammars`. This worked nicely and I had basic XML syntax coloring.

However, now I can't remove the grammar anymore. The `Remove...` button is disabled when I select the grammar, and just deleting the grammar file leads to a ton of exceptions in various places...

TM4E version: 0.1.0.201709192210

cmoine:
I can provide a PR for that if you want + A button "Edit" that would be convenient I think.

mickaelistria:
A PR would be very welcome!

angelozerr:
> I can provide a PR for that if you want + A button "Edit" that would be convenient I think.

It should be very cool! If I remember the problem with remove grammar is when a grammar is used in an another grammar (embed grammar), it should reset the other grammar which are bound to the grammar which is deleted. But it's perhaps a complex case and we can start just by deleting a grammar.

cmoine:
I will first make something simple that works. The condition for button enability, and the condition for processing is inverted, which ends up with a NO-OP :D

cmoine:
I send the pull request but Travis is complaining:
```
$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"
Ignoring license option: BCL -- using GPLv2+CE by default
install-jdk.sh 2019-07-17
Expected feature release number in range of 9 to 14, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .
```
Any idea ?

mickaelistria:
Fixed with #218. Please try https://download.eclipse.org/tm4e/snapshots/ to verify.

cmoine:
Perfect thank you @mickaelistria

PR #218, opened 2019-08-30
#149 Fix the Remove Grammar button

cmoine:
the condition for button enability and the condition for processing is inverted, which ends up with a NO-OP


mickaelistria:
Thank you!

Issue #202, opened 2019-04-07
Can't get Generic editor to use provided Textmate grammars


fbricon:
Using tm4e  0.3.2.201901202056, I installed the Kotlin TM grammar from https://github.com/mathiasfrohlich/vscode-kotlin/blob/master/syntaxes/Kotlin.tmLanguage, and the Groovy grammar from https://github.com/textmate/groovy.tmbundle/blob/master/Syntaxes/Groovy.tmLanguage. Highlighting works in the tm4e preview panel, but not when opening the files with the generic editor. See for kotlin:
![Apr-07-2019 11-43-11](https://user-images.githubusercontent.com/148698/55686251-caf50880-592c-11e9-90f0-ffe13e2c3b8c.gif)

mickaelistria:
Did you enable the TMPresentationReconcilier for the expected content-types in generic editor? https://github.com/eclipse/aCute/blob/master/org.eclipse.acute/plugin.xml#L72
Did you associate the TextMate grammar/scope with desired content-types? https://github.com/eclipse/aCute/blob/master/org.eclipse.acute/plugin.xml#L59

mickaelistria:
See https://github.com/eclipse/tm4e/wiki/UI

fbricon:
As a simple user, my expectation is that, once tm4e is installed and I provide textmate grammars (I wish a bunch of those were already provided OOTB, or I could discover and install them through some UI at least, but that's another story), it would just work in the generic editor. Anything that requires me to write plugin.xml configuration is a complete no-go, for *me*. So it seems I misunderstood the goal and scope of the project, which is fine, but maybe the readme ought to be clearer.

mickaelistria:
Better workflow for user addition of TextMate grammars would be a nice to have, but it's not really considered as a big factor of success for TM4E nor the IDE. Having only syntax highlighting isn't usually the value we aim to provide, nor the value most users expect in an IDE. If you want Kotlin or Groovy support, then it's better to install the full Kotlin and Groovy plugins. Whether they decide to use TM4E or something else is up to them.
At the moment, TM4E targets plugin developers, not really end-users. This could indeed be improved.

Please re-title the bug to better capture it as an enhancement request.

angelozerr:
> As a simple user, my expectation is that, once tm4e is installed and I provide textmate grammars

[...]

PR #219, opened 2019-08-30
#202 Fix the file extension lookup

cmoine:


org.eclipse.tm4e.registry/src/main/java/org/eclipse/tm4e/registry/internal/AbstractGrammarRegistryManager.java
```diff
@@ -122,7 +122,8 @@ public IGrammar getGrammarForFileType(String fileType) {
 	 * 
 	 * @return
 	 */
-	public IGrammarDefinition[] getDefinitions() {
+	@Override
+    public IGrammarDefinition[] getDefinitions() {
```
mickaelistria: Incorrect indentation here. Please mimic what's on previous line.

org.eclipse.tm4e.registry/src/main/java/org/eclipse/tm4e/registry/internal/AbstractGrammarRegistryManager.java
```diff
@@ -173,7 +174,8 @@ public IGrammarDefinition getDefinition(String scopeName) {
 	 * @return list of scope names to inject for the given
 	 *         <code>scopeName</code> and null otheriwse.
 	 */
-	public Collection<String> getInjections(String scopeName) {
+	@Override
+    public Collection<String> getInjections(String scopeName) {
```
mickaelistria: indentation

org.eclipse.tm4e.registry/src/main/java/org/eclipse/tm4e/registry/internal/AbstractGrammarRegistryManager.java
```diff
@@ -198,7 +200,8 @@ public String getScopeNameForContentType(String contentTypeId) {
 		return pluginCache.getScopeNameForContentType(contentTypeId);
 	}
 
-	public String[] getContentTypesForScope(String scopeName) {
+	@Override
+    public String[] getContentTypesForScope(String scopeName) {
```
mickaelistria: indentation.

org.eclipse.tm4e.registry/src/main/java/org/eclipse/tm4e/registry/internal/AbstractGrammarRegistryManager.java
```diff
@@ -122,7 +122,8 @@ public IGrammar getGrammarForFileType(String fileType) {
 	 *
```
mickaelistria: Overall, this file doesn't have any payload change, it'd be better to just not change it at all.

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/text/TMPresentationReconciler.java
```diff
@@ -408,7 +408,7 @@ private IGrammar findGrammar(ContentTypeInfo info) {
 				// Discover the well grammar from the filetype
 				String fileName = info.getFileName();
 				if (fileName != null) {
-					String fileType = new Path(fileName).getFileExtension();
+					String fileType = "." /* #202 */ + new Path(fileName).getFileExtension();
```
mickaelistria: I don't think this line should be changed. Implementation of `getGrammarForFileType` should be fixed to receive `sh` or `.sh` and return the same result.
Maybe it would even be better to have the IGrammarRegistryManager have a getGrammarForFileName(...) instead that would take care of the logic of extracting the right subset of the file name.

cmoine: I quite agree on this, since I believe that some tmLanguage file define sometimes ".sh", sometimes "sh".

mickaelistria:
Thank you!

PR #222, opened 2019-09-04
Allow setTheme to be called several times in a row

### LINKED ISSUE #220: Switch to theme only works once for open editor
Hello,

Using the "Switch to Theme..." menu works the first time but fails on subsequent attempts until the editor is closed and reopened again.

I did a bit of debugging and discovered that this behaviour is caused by this [particular if condition](https://github.com/eclipse/tm4e/blob/5c855d43cf10e3681b8b0c52df71fcd5e898eec7/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/text/TMPresentationReconciler.java#L170). The `forcedTheme` handling was apparently introduced in a390c266c343bfe69519c90d3a6105da9ed5f6fd. The commit message isn't very explanatory, does anyone have more insight on why was this added? Is it still needed now?

Cheers,

Pyves

PyvesB:
Fixes  #220. I haven't noticed any negative side-effects whilst testing, both the Switch to Theme menu and the preferences windows switching now work several times in a row.

Signed-off-by: Pierre-Yves B. <PyvesDev@gmail.com>


mickaelistria:
Thank you!

PR #247, opened 2020-06-04
Fixes to avoid NPEs

briandealwis:
Two simple fixes for NPEs observed in practice:

  - I periodically encountered situations where the state was null in the Tokenizer with the Go textmate grammars at https://github.com/syscrusher/golang.tmbundle/
  - It's possible for a document to not have a content-type


mickaelistria:
See CI log, the version of languageconfiguration plugin must be bumped by +0.0.1

mickaelistria:
Thanks @briandealwis !

Issue #261, opened 2020-11-03
NullPointerException when using custom theme file


j-ulrich:
I am trying to use a custom theme (.css) file. But after adding it to the TextMate themes list, I get a NullPointerException whenever the theme is used (e.g. clicking the theme in the list, setting the theme for a grammar, opening a file after the theme has been set etc.):

```
eclipse.buildId=4.17.0.I20200902-1800
java.version=14.0.2
java.vendor=Oracle Corporation
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=de_DE

org.eclipse.jface
Error
Tue Nov 03 10:22:20 CET 2020
Problems occurred when invoking code from plug-in: "org.eclipse.jface".

java.lang.NullPointerException
	at org.eclipse.tm4e.ui.themes.css.CSSTokenProvider.getEditorBackground(CSSTokenProvider.java:89)
	at org.eclipse.tm4e.ui.themes.Theme.getEditorBackground(Theme.java:97)
	at org.eclipse.tm4e.ui.themes.Theme.initializeViewerColors(Theme.java:160)
	at org.eclipse.tm4e.ui.internal.widgets.TMViewer.setTheme(TMViewer.java:74)
	at org.eclipse.tm4e.ui.internal.preferences.ThemePreferencePage.preview(ThemePreferencePage.java:337)
	at org.eclipse.tm4e.ui.internal.preferences.ThemePreferencePage.lambda$0(ThemePreferencePage.java:189)
	at org.eclipse.jface.viewers.Viewer$1.run(Viewer.java:151)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:174)
	at org.eclipse.jface.viewers.Viewer.fireSelectionChanged(Viewer.java:148)
	at org.eclipse.jface.viewers.StructuredViewer.updateSelection(StructuredViewer.java:2132)
	at org.eclipse.jface.viewers.StructuredViewer.handleSelect(StructuredViewer.java:1178)
	at org.eclipse.jface.viewers.StructuredViewer$4.widgetSelected(StructuredViewer.java:1207)
	at org.eclipse.jface.util.OpenStrategy.fireSelectionEvent(OpenStrategy.java:242)
	at org.eclipse.jface.util.OpenStrategy.access$4(OpenStrategy.java:237)
[...]

PR #263, opened 2020-11-11
Adds exception messages

j-ulrich:
Relates to #261


j-ulrich:
This is untested, though, because I don't have a proper setup available here.

mickaelistria:
Thanks!

PR #288, opened 2021-10-21
Support locations outside a local file system

rubenporras:
Support locations outside a local file system by giving
buffer.getContentType() precedence.

In addition fix a NPE in getContents() if the location is null (always
the case for remote resources).


rubenporras:
Oh no, yet another place with the same problem :) 

I have quickly debugged the code and it works for me but you are right, to be complete, it should also use buffer.getContentType(), I send a patchset for that as well.

Thanks

PR #299, opened 2021-12-03
Minor code cleanup

sebthom:

PR #300, opened 2021-12-04
add GitHub Actions build

sebthom:
This adds commits on top of PR #299 which should be merged first.


mickaelistria:
Let's give a try!

sebthom:
Thanks. The GHA builds start faster and need less time to run, thus contributors get faster feedback.

PR #302, opened 2021-12-04
minor code cleanup

sebthom:
I ran the Eclipse Code Cleanup action with a few uncritical rules.

Apparently a few files where committed with wrong line endings before. So better do a diff ignoring whitespaces using `?w=1`: https://github.com/eclipse/tm4e/pull/302/files?w=1


mickaelistria:
Thanks!

PR #303, opened 2021-12-04
Improve .gitignore settings

sebthom:


mickaelistria:
.project and .settings should be committed

sebthom:
@mickaelistria .project is currently commit for all except the root project. no idea why. I can add it.
currently there are no .settings folders committed. I can add them if you want.

mickaelistria:
Yes, please add .settings,.

sebthom:
I changed gitignore so that .settings/org.eclipse.core.resoures.prefs, .settings/org.eclipse.jdt.core.prefs, .settings/org.eclipse.pde.core.prefs are managed by git. I am leaving out other files in .settings as other plugins and their settings may vary between user/Eclipse installation which can result in conflicts if different preferences are used.

mickaelistria:
Please keep all settings. Those are project configuration that are expected to be portable. What settings did you find isn't portable?

sebthom:
Ever custom plugin one installs may potentially write into the settings folder. E. g. AnyEdit, mark down editors etc. These settings may be very user specific. The only ones that are really relevant to be checked in are the once that influence the CI build, i. e. the Eclipse compiler settings picked up by the tycho compiler. not custom UI settings. I for example find it extremely annoying when I checkout a project from git and directly all (child) projects are marked as modified in Eclipse because the complete settings folder is managed by git. Then if you have users with different preferences for non-build relevant settings you constantly get merge conflicts.

mickaelistria:
> These settings may be very user specific

Not really. User specific settings go to the workspace, project settings are specific to the project.

> The only ones that are really relevant to be checked in are the once that influence the CI build, i. e. the Eclipse compiler settings picked up by the tycho compiler. not custom UI settings.

Sharing the settings is not only for CI, but also, and mainly for other contributors: things like code style, save actions, default profiles for the IDE... are all stored in settings and having them pre-set for contributors is a way to make it easier to get started with the project and avoid annoying comments in reviews about details that could have been prevented by shared settings.

> I for example find it extremely annoying when I checkout a project from git and directly all (child) projects are marked as modified in Eclipse because the complete settings folder is managed by git.

That's not supposed to happen. What makes the children settings marked as modified upon import? What changes are actually happening?

> Then if you have users with different preferences for non-build relevant settings you constantly get merge conflicts.

The project preferences are not supposed to contain user preferences, but some configuration of the project, that should apply to all users.

mickaelistria:
Let's merge anyway despite our disagreement.

PR #309, opened 2022-01-07
Evaluate command as enter when text equals delimiter

bknaysi:
## Explanation of the bug
Hello! I found a bug in the `TextUtils.isEnter(IDocument d, DocumentCommand c)` method. I put an explanation of the bug in the description of the commit. TLDR - the function identifies document commands with a 0 length selection and clipboard text ending in a delimiter as an enter key press.

## Merging the pull request
We currently use the `0.1.0` release in our product. Since there isn't a maintenance branch for the `0.1.0` release, I opened this pull request against the `master` branch. I believe a maintainer will need to create a new branch from the `0.1.0` tag, and change this pull request's base to that new branch. Lastly, they would commit any version updates, create a `0.1.1` tag, and run a build. All of this assuming you agree with my change!

## How we consume builds
Our build pulls from the [tm4e software repository](https://archive.eclipse.org/tm4e/releases/) downloads. Ideally your pipeline publishes the new version there. Thanks for the help and active maintenance of this project!


mickaelistria:
This seems like a valid patch, but this needs a manual rebase against master first. Please take care of it.

bknaysi:
Thanks for the reply! I'm closing this because the fix exists in the `0.2.0` release; I forgot to check.

PR #318, opened 2022-02-21
Fixing some minor issues

sebthom:
I addressed a few issues reported by the Eclipse compiler. Each fix is in a separate commit for easier reviewing.


sebthom:
@mickaelistria the unit test failures seem not to be related to the changes, as they already occurred in previous builds.

akurtakov:
Thanks for the patches. Tests succeed on Jenkins so it's fine.

PR #329, opened 2022-04-11
Cleanup JSONPlistParserTest and XMLPlistParserTest

sebthom:
- use try-with-resources to fix resource leak
- use proper name for test methods
- add some assertions to test methods
- remove duplicate test plist resources


akurtakov:
Thanks for the patch!

PR #330, opened 2022-04-12
Support loading of TextMate Grammars in YAML format

sebthom:


org.eclipse.tm4e.core/META-INF/MANIFEST.MF
```diff
@@ -9,7 +9,8 @@ Require-Bundle: org.apache.batik.css;bundle-version="1.9.1";resolution:=optional
  org.apache.batik.util;bundle-version="1.9.1";resolution:=optional,
  com.google.gson;resolution:=optional,
  org.jcodings;bundle-version="1.0.18",
- org.joni;bundle-version="2.1.11"
+ org.joni;bundle-version="2.1.11",
+ org.yaml.snakeyaml;bundle-version="1.27.0"
```
angelozerr, line 13: I wonder if there are a lot of textmate grammar written in YAML? If no, perhaps we could set this bundle as optional?

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/parser/yaml/YamlPListParser.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.InputStream;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import org.eclipse.tm4e.core.internal.parser.PList;
+import org.xml.sax.SAXException;
+import org.yaml.snakeyaml.Yaml;
+
+public class YamlPListParser<T> {
```
angelozerr, line 24: Add Javadoc

org.eclipse.tm4e.core/META-INF/MANIFEST.MF
sebthom: VSCode supports it so I think it would be good to support it in tm4e out-of-the-box too. Writing grammars in YAML has the advantage over JSON that it's more compact and supports aliases and comments.

mickaelistria:
Thanks a lot!

org.eclipse.tm4e.core/META-INF/MANIFEST.MF
angelozerr: OK thanks for your feedback. With JSON you can too write  coments (jsonc)

It makes sense to have yaml parser  support.

PR #331, opened 2022-04-12
Rename PListFormatsTest to GrammarReaderTest

sebthom:
Renames/moves the `internal.grammar.parser.PListFormatsTest` to `internal.grammar.reader.GrammarReaderTest`, since the test mainly uses the GrammarReader from this package.
I also added some assertions and JavaDoc.


mickaelistria:
some test is now failing

sebthom:
@mickaelistria I removed the hashCode/equals override again I added to the `Raw` class. I was not aware that of the fact that the fileTypes field is lazy initialized and basically only used as a cache. So when calling `IRawGrammar.getFileTypes()` the first time it would chang the hashcode and comparing two IRawGrammer objects where on one `getFileTypes()` was called were not equal anymore.

mickaelistria:
Looks good to me and tests, thanks.

PR #334, opened 2022-04-12
Fix incomplete listeners synchronization in TMPresentationReconciler

sebthom:
Doing unsynchronized null-checks has the known issues of Double Checked
Locking, where the current field value may not be visible in other
threads. Even without that issue concurrent calls to
addTMPresentationReconcilerListener() can result in race-conditions
because of the lazy initialized unsynchronized assignment of the
listeners variable.


mickaelistria:
Thanks!

PR #335, opened 2022-04-12
Use Guava String Splitter

sebthom:
The split methods avoid unnecessary temporary object allocation and fast-exit for lower latency.

These split methods are reliable and have 100% code coverage.


sebthom:
Some context: Currently e.g. `scope.split("[.]")` is used in the code to split a string by a single dot character. Despite the fact that the JVM's split method has a separate code path for splitting my single characters it does not work when splitting by a single dot since the regex pattern required is longer than one character and thus the split methods optimized non-regex branch is not used.

In a wider context, I am aiming to incrementally improve the general TM4E/LSP4E performance/responsiveness by reducing object allocation/GC pressure. This is only one part of it.

mickaelistria:
OK. But aren't there some common APIs (apache-commons, Guava, ....) which provide the same functionality and that we could just reuse?

sebthom:
That means introducing another dependency. I don't know if you would want that.

I checked the Apache Commons `StringUtils.split(String,char)` implementation. The implementation has a subtle difference, it does not return a list with a single-empty string entry in case the input string was an empty string, so it works different from `String.split()`. As far as I understand the TM4E code base however expects this behaviour, so the Apache Commons StringUtils could not be used as a drop-in replacement.
Apache Commons `StringUtils` also does not provide a split method that returns a List, so if you need a list you have to go via a temporary String Array. Which means in total 2 Lists (one internal in the split method) and one String array will be instantiated on each split operation.

mickaelistria:
> That means introducing another dependency. I don't know if you would want that.

we usually prefer another dependency than more duplicating a functionality and adding code to maintain.
So if we can use StringUtils.split from Apache, it would IMO be best.

> That means introducing another dependency. I don't know if you would want that.

Can you please point to a piece of code where there is a requirement on `""split(...) returns [""]`? I'm not sure it's a functional requirement, but instead could be some workarounds particularly for this story of `String.split()` not returning `[]` as some could expect.

sebthom:
I changed the PR to replace regex based splits with Guava Splitter. Guava Splitter by default behaves like String.split on empty strings meaning it returns a list with a single entry containing the original (empty) string.

mickaelistria:
Thanks!

PR #337, opened 2022-04-13
Replace if-string-equals statements with switch statements

sebthom:


mickaelistria:
1 not so related question (as I see many opportunities for switch-expressions), what is the minimal version of Java you need to run with? Maybe it would be worth moving to Java 17 as BREE.

sebthom:
Java 11 which is still the minimal required version by eclipse sdk according https://wiki.eclipse.org/Eclipse/Installation
I am not sure if requiring a higher JRE for tm4e would be a good idea atm.

mickaelistria:
Since JustJ ships Java 11 and Java 17 and is easy to use from a technical and legal POV, and does serve most users of Eclipse IDE, I don't think we should feel forced to stay against older Java, unless it makes sense to the project committers to keep support for older Java. That's why I'm asking there.

sebthom:
Maybe a survey via the mailing list would make sense. This also impacts lsp4e, right?

PR #338, opened 2022-04-13
Improve logging performance

sebthom:


sebthom:
That would be a bigger change. Also I don't know how to deal with: https://github.com/eclipse/tm4e/blob/e67293312a96813569d56cc5abb035257f542b9a/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/TMUIPlugin.java#L62-L91

mickaelistria:
Usually it's an anti-pattern to add conditions about logging and debugging in business code. So even if it's more work, I think it's profitable to progressively move to System.Logger as much as possible.
For specific handling, I think this has to do with providing a different LoggerFinder. However, as System.Logger default to java.util.logging.Logger (unless another System.Logger implementation is provided), then the code you're linking here should still work by default: System.Logger would call this java.util.logging.Logger and this java.util.logging.Logger would then re-route to Eclipse logs.

sebthom:
@mickaelistria I changed to java.lang.System.Logger

mickaelistria:
Perfect, thanks!

PR #345, opened 2022-04-16
Fix warning 'Statement unnecessarily nested within else clause'

sebthom:


sebthom:
For easier review compare ignoring whitespace changes: https://github.com/eclipse/tm4e/pull/345/files?w=1

PR #366, opened 2022-04-22
Make XML parser invulnerable to XXE attacks

sebthom:
See https://rules.sonarsource.com/java/RSPEC-2755


sebthom:
@mickaelistria @akurtakov Do you think this is safe from a functional point of view? Could this have of any side effects?

akurtakov:
I see plist files have doctype http://www.apple.com/DTDs/PropertyList-1.0.dtd . This change would make this dtd to not be loaded so we may end up with parsing error later rather than earlier. I'm not that comfortable with this change for that.

mickaelistria:
I don't think DocType are necessary for the parser, from a functional POV. They may be useful in order to produce validation and other stuff that is more relevant at dev-time, but in TM4E context, this wouldn't apply.

angelozerr:
I agree with @mickaelistria except if we wish to validate the PList when it is loaded.

PR #367, opened 2022-04-22
Make gson dependency non-optional

sebthom:
The gson-base JSON parser is not implemented or referenced in a dynamic/optional way but is hard referenced in the code at in the [GrammarReader](https://github.com/eclipse/tm4e/blob/5a3c7ba072c3ee6988425e8b6ca5ba3e67cfde7e/org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/internal/grammar/reader/GrammarReader.java#L80). Thus gson is not an optional dependency.


sebthom:
@mickaelistria @akurtakov any objections?

PR #369, opened 2022-04-23
Enable annotation-based null analysis for some packages

sebthom:


org.eclipse.tm4e.core/META-INF/MANIFEST.MF
```diff
@@ -11,7 +11,8 @@ Require-Bundle: org.apache.batik.css;bundle-version="1.9.1";resolution:=optional
  com.google.guava;bundle-version="30.1.0",
  org.jcodings;bundle-version="1.0.57",
  org.joni;bundle-version="2.1.43",
- org.yaml.snakeyaml;bundle-version="1.27.0"
+ org.yaml.snakeyaml;bundle-version="1.27.0",
+ org.eclipse.jdt.annotation;bundle-version="2.2.600";resolution:=optional
```
akurtakov, line 15: Null annotations are not retained at runtime so no need for OSGi container to know/care about them. (Here) [https://codeiseasy.wordpress.com/2013/03/08/tycho-and-jdt-null-analysis/] is a way to do it as build time only.

sebthom: I don't understand your comment. I marked the bundle as optional, same as it was done a few years ago in the tm4e.ui module at [https://github.com/eclipse/tm4e/blob/master/org.eclipse.tm4e.ui/META-INF/MANIFEST.MF#L21](https://github.com/eclipse/tm4e/commit/a4c43ea28ac456d3bcd68485b074921f310edd7a#diff-975e21d444520f30c3d426a50429bd8dceeaf5d75d0fb5080fe983d7754c3fa9).

If I don't reference the bundle I cannot import `@Nullable` while editing the files in Eclipse.

sebthom: Ah...build.properties instead of MANIFEST.MF, got it. Thanks!

akurtakov: I'll change ui bundle myself.

akurtakov: Use the same approach as https://github.com/eclipse/tm4e/commit/a7f5c8283b4ec13e36159f886d9276b7f8c112e1

sebthom: Done

akurtakov:
if you feel comfortable with the change please merge it as it's already quite big. I would not have the time to look at it.

PR #372, opened 2022-04-27
Remove annotations from OSGi 

akurtakov:


target-platform/tm4e-target.target
```diff
@@ -21,18 +21,18 @@
             <unit id="org.junit.jupiter.api" version="0.0.0"/>
             <unit id="org.eclipse.jdt.junit5.runtime" version="0.0.0"/>
             <unit id="org.eclipse.ui.trace" version="0.0.0"/>
-            <repository location="https://download.eclipse.org/eclipse/updates/4.22/"/>
+            <repository location="https://download.eclipse.org/eclipse/updates/4.23/"/>
```
sebthom, line 24: Because of https://bugs.eclipse.org/bugs/show_bug.cgi?id=579347 / https://bugs.eclipse.org/bugs/show_bug.cgi?id=579153 which makes Eclipse 4.23 unusable in a lot of situations I would appreciate if we could stay still target 4.22 for now until 4.24 is released where this is solved.

akurtakov: Noted. Will remove this commit.

PR #377, opened 2022-05-02
Move StackElement to internal package and add public IStackElement interface

sebthom:
StackElement references internal classes in it's methods signatures.


sebthom:
@mickaelistria @angelozerr @akurtakov any objections?

mickaelistria:
No objection, it is consistent with other types of the grammar model.

PR #378, opened 2022-05-02
Refactor theme package

sebthom:


org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/theme/Theme.java
```diff
@@ -226,32 +222,39 @@ public static Theme resolveParsedThemeRules(final List<ParsedThemeRule> parsedTh
...
+
+		if (RGB.matcher(hex).matches()) {
+			// #rgb
+			return true;
+		}
+
+		if (RGBA.matcher(hex).matches()) {
+			// #rgba
+			return true;
+		}
```
mickaelistria, line 248:
```suggestion
		return RRGGBB.matcher(hex).matches() || RRGGBBAA.matcher(hex).matches() || RGB.matcher(hex).matches() || RGBA.matcher(hex).matches();
```

sebthom: The current code base tries to stay close to the upstream typescript project. so if bugs are solved there they can be easily ported to tm4e by visual comparison of the code bases. The equivalent code in vscode is: https://github.com/microsoft/vscode-textmate/blob/9157c7f869219dbaf9a5a5607f099c00fe694a29/src/utils.ts#L132-L154

PR #267, opened 2020-12-17
Issue #266 - [WIP] Reproducer + come cleanup

mickaelistria:
Signed-off-by: Mickael Istria <mistria@redhat.com>

Issue #266, opened 2020-12-17
multi-line yaml values not rendered properly


mickaelistria:
See https://github.com/eclipse/wildwebdeveloper/issues/563

sebthom:
This issue might be fixed in the meantime. I ported a few fixes from the vscode textmate project.

PR #381, opened 2022-05-04
Add multiline YAML test

sebthom:
Test case taken from #266 / #267


sebthom:
@mickaelistria No we don't. I can also just include the YAML test case and leave the tokenizeText methods out.

mickaelistria:
I don't think there is any benefit in adding code we don't use, it only adds maintenance and doesn't deliver value. So yes, I'd prefer we leave the tokenizeTest methods out as long as we don't use them.

PR #386, opened 2022-05-06
Use Eclipse Collections IntArrayList

sebthom:


mickaelistria:
I don't see a significant gain with this; while it looks like it actually require to add code and increase the size of the dependency tree. What particular benefit makes you think it's worth it?

mickaelistria:
I've gone a bit through the code and I think before adding new types, we should focus on removing content. Many pieces of code use some ArrayList and the MoreCollections class. Actually, all that could be greatly simplified by using some LinkedList which has some pop(), last() and other methods. LinkedList actually maps better the need of a parser which usually consists of FIFO or LIFO and for which we don't often need to address items by index.

PR #389, opened 2022-05-09
Add org.eclipse.tm4e.core.benchmarks

sebthom:
This PR adds a test-plugin containing a JMH based benchmark of Grammar#tokenizeLine.
The benchmark is not executed during the regular build but can be launched using the provided run-benchmark.sh / run-benchmark.cmd scripts.


mickaelistria:
That's very interesting. I'm not much familiar with JMH and benchmarks. Can you please add a README.md file or something like that in the module to describe how to run it and how to interpret the results and whatever info that could help?

PR #390, opened 2022-05-09
add org.eclipse.tm4e.core.benchmarks module with GrammarBenchmark

sebthom:


sebthom:
@mickaelistria I now created a benchmark from scratch. It has two nice side effects (besides not having the licensing issue): the benchmark can now also provide some indications regarding memory consumption and it can for a quick test be directly executed from within Eclipse.

I also added a readme with some explanations https://github.com/eclipse/tm4e/blob/43373077e147ff936e097617c7669685a1993327/org.eclipse.tm4e.core.benchmarks/README.md

mickaelistria:
Does it have to be an OSGi bundle? Couldn't it just be a plain Java module? Also, would it be possible/better to actually include this code in some `src/benchmark/java` folder of org.eclipse.tm4e.core directly?

sebthom:
If it is a separate module it has to be an OSGI module, otherwise (transitive) dependency resolution does not work. But I can make another PR where the benchmark is directly in the tm4e.core module.

mickaelistria:
No need to change it. those were naive questions, let's just do what you think is best.

PR #391, opened 2022-05-10
Add GrammarBenchmark

sebthom:


sebthom:
@mickaelistria This is an alternative PR where I added the GrammarBenchmark directly to the core module. Please check if the location of the files (readme, shell wrapper, etc) is ok and give it a spin if the benchmark values make sense to you.

mickaelistria (approved):
Thanks. Yes, it seems all right to me. Feel free to merge if you don't see any reason to keep it pending.

Issue #387, opened 2022-05-07
NoClassDefFoundError when open pom.xml


Hawkeyes0:
stack trace:

```
java.lang.NoClassDefFoundError: org/eclipse/tm4e/ui/utils/ContentTypeHelper
	at org.eclipse.tm4e.languageconfiguration.LanguageConfigurationCharacterPairMatcher.findContentTypes(LanguageConfigurationCharacterPairMatcher.java:134)
	at org.eclipse.tm4e.languageconfiguration.LanguageConfigurationCharacterPairMatcher.getMatcher(LanguageConfigurationCharacterPairMatcher.java:113)
	at org.eclipse.tm4e.languageconfiguration.LanguageConfigurationCharacterPairMatcher.match(LanguageConfigurationCharacterPairMatcher.java:47)
	at org.eclipse.jface.text.source.MatchingCharacterPainter.paint(MatchingCharacterPainter.java:313)
	at org.eclipse.jface.text.PaintManager.addPainter(PaintManager.java:207)
	at org.eclipse.jface.text.TextViewer.addPainter(TextViewer.java:5175)
	at org.eclipse.ui.texteditor.SourceViewerDecorationSupport.showMatchingCharacters(SourceViewerDecorationSupport.java:679)
	at org.eclipse.ui.texteditor.SourceViewerDecorationSupport.updateTextDecorations(SourceViewerDecorationSupport.java:283)
	at org.eclipse.ui.texteditor.SourceViewerDecorationSupport.install(SourceViewerDecorationSupport.java:269)
	at org.eclipse.ui.texteditor.AbstractDecoratedTextEditor.createPartControl(AbstractDecoratedTextEditor.java:456)
	at org.eclipse.ui.internal.genericeditor.ExtensionBasedTextEditor.createPartControl(ExtensionBasedTextEditor.java:166)
	at org.eclipse.ui.part.MultiPageEditorPart.addPage(MultiPageEditorPart.java:227)
	at org.eclipse.ui.forms.editor.FormEditor.addPage(FormEditor.java:331)
	at org.eclipse.ui.part.MultiPageEditorPart.addPage(MultiPageEditorPart.java:203)
	at org.eclipse.ui.forms.editor.FormEditor.addPage(FormEditor.java:313)
	at org.eclipse.m2e.editor.pom.MavenPomEditor.addFallbackEditors(MavenPomEditor.java:343)
[...]

PR #393, opened 2022-05-10
Some minor changes

sebthom:


sebthom:
@mickaelistria I added minimum version constraints for tm4e plugins referencing each other because of the refactoring. To prevent partial updates and issues like #387

PR #396, opened 2022-05-12
Fix build

sebthom:


target-platform/tm4e-target.target
```diff
@@ -21,15 +21,28 @@
...
+            <dependencies>
+                <dependency>
+                    <groupId>com.google.code.gson</groupId>
+                    <artifactId>gson</artifactId>
+                    <version>2.9.0</version>
+                    <type>jar</type>
+                </dependency>
+            </dependencies>
+            <instructions><![CDATA[
+Bundle-Name:           Bundle derived from maven artifact ${mvnGroupId}:${mvnArtifactId}:${mvnVersion}
```
akurtakov, line 35: Why is this needed? I just checked and https://repo1.maven.org/maven2/com/google/code/gson/gson/2.9.0/gson-2.9.0.jar is osgi artifact already?

sebthom: Then it is not. I didn't know.

PR #400, opened 2022-05-17
Adapt changes in upstream project

sebthom:
A large (and good!) code refactoring took place in vscode-textmate. This PR adapts the changes to tm4e.


mickaelistria:
This is a huge change, I don't think I'll really do a full review. I see some more tests were added and existing tests still seems to work; code does look slightly better and I trust if you say that it's actually better. So if you feel confident merging this, feel free to do so!

PR #402, opened 2022-05-20
Replace non-functional stopAtOffset tokenize parameter with timeLimit

sebthom:
The `stopAtOffset` parameter of `ITokenizationSupport#tokenize` was not implemented and the **tm4e.ui** plugin was not using it. So I replaced it with the `timeLimit` parameter which I already added to `IGrammar#tokenizeLine` according to the vscode-textmate project.

When this PR is merged I am going to work on the TMModel implementation. Having timeLimit support in ITokenizationSupport will allow to greatly simplify the TMModel class.

### DIFF
--- a/org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/ITokenizationSupport.java
+++ b/org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/ITokenizationSupport.java
@@ -16,6 +16,8 @@
  */
 package org.eclipse.tm4e.core.model;
 
+import java.time.Duration;
+
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.tm4e.core.grammar.IStateStack;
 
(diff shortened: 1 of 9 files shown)


org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/ITokenizationSupport.java
```diff
@@ -29,9 +29,10 @@ public interface ITokenizationSupport {
...
 	TokenizationResult tokenize(String line, @Nullable IStateStack state);
 
-	// add offsetDelta to each of the returned indices
-	// stop tokenizing at absolute value stopAtOffset (i.e. stream.pos() +
-	// offsetDelta > stopAtOffset)
-	TokenizationResult tokenize(String line, @Nullable IStateStack state, Integer offsetDelta, Integer stopAtOffset);
-
+	/**
+	 * @param offsetDelta adds offsetDelta to each of the returned indices
+	 * @param timeLimit duration in milliseconds after which tokenization is stopped
```
mickaelistria, line 34: What about using a Duration type directly?

sebthom: Great idea! Just changed it.

mickaelistria (approved):
Let's merge.

PR #403, opened 2022-05-24
Refactor TMModel line token update and DocumentModelLines

sebthom:
I refactored/simplified the way the TMModel handles line updates. Still seems to work as expected while needing less LoCs and I believe being easier to follow. Also it now makes use of the new Grammar's timeLimit feature.

Since this is a substantial change of the current logic. It would be great if you could also try it out before merging.


org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/ModelLine.java
```diff
@@ -36,6 +36,10 @@ public List<TMToken> getTokens() {
 		return tokens;
 	}
 
+	public boolean isValid() {
+		return !isInvalid;
+	}
+
 	public boolean isInvalid() {
```
mickaelistria, line 43: Should we remove `isInvalid` then?

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/ModelLine.java
```diff
@@ -36,6 +36,10 @@ public List<TMToken> getTokens() {
 		return tokens;
 	}
 
+	public boolean isValid() {
```
mickaelistria, line 39: I actually don't like much the name `isValid` here. Indeed, in such an approach, being "invalid" is semantically precise as "state is not ready", it's actually not the opposite of valid. So if we can avoid adding this method, I think it keeps things clearer for consumers.

mickaelistria (commented):
I don't have much knowledge to judge this change but trust it is good ;)
I just did add an minor comment about the new method, but this is not a blocker.

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/ModelLine.java
sebthom: I also thought about this. I think what invalid means is that the contained information is out of date / out of sync with the actual line in the editor. So we could change it to something like isOutdated / isSynced / isUpToDate, etc. any preference?

angelozerr:
What about isDirty?

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/ModelLine.java
mickaelistria: We already have `isInvalid()` which is pretty explicit. Do we need a negated variant at all? If yes, then I thing `isUpToDate` is the closest meaning.

sebthom: I have no hard feelings for either getter. I can remove one. I just found `line.isValid()` nicer to use than having to use `!line.isInvalid()` which is a double negation.

mickaelistria: Our software world abuse the "invalid" term so it's not really a negation but a precise description of an actual state unlike "valid" ;) But I'm totally fine with @angelozerr suggestion of renaming `isInvalid` to `isDirty` if it helps.
My only concern is more that I don't like `isValid`.

sebthom:
I decided not to change the name of `ModelLine.isInvalid` to anything else for now. However now I also did some changes to the `org.eclipse.tm4e.ui.internal.model` to reduce the number of reparsings and thus editor refreshes/repaintings to 1 on multi-line edits. Until now for each multi-line change two separate line invalidation events were send to the Tokenizer thread. This should speed up working on larger documents.

mickaelistria:
To be honest, I'm now much less familiar with this code than you are. I had a quick look at the general change here and think it's indeed a good one; so if you're confident enough with it and since tests are passing, feel free to merge.

Issue #305, opened 2021-12-15
smarter or more granular auto closing of certain characters


jcompagner:
in the preferences we have:

![image](https://user-images.githubusercontent.com/3274268/146173122-0d1c5daf-389f-4fae-8c42-3dae14c8bd88.png)

Which is quite broad, you can't say for `"` I want to auto close but not for `{`.
It's all or nothing.
Also it is called "enabling auto close brackets" but its not always a **bracket** right? 

But I would like that it would be a bit more smarter, because in the Typescript editor I really just want to turn it all off.
But in the Java editor I am not so bothered by that, it seems they know a bit better when or when not to do it.

As an example:
```typescript
if (true) statement;
```
If i now want to make this:
```typescript
if (true) { statement; }
```
In Java it is easy, I can do `{` and it won't insert a `}` right away and I can move to the end and place a `}` myself.
But in the typescript editor it right away inserts `{}` so I have to delete it again myself.

Same for example for strings.
In Java if you just type `"` and there is nothing else around it it will make `""` and that's fine.
But if I type it just just in front of another char so I have:
```typescript
a
```
and I place the cursor just before the `a` and type `"` I get:
```typescript
"a
```
not a double `""a` that I get in ts editor.

Those are the things that make the auto close a bit smarter so it's not annoying that you just want to turn it off.

sebthom:
@mickaelistria do you think it makes sense we change it so brackets are only automatically closed if inserted at the end of a line of  line? I don't think we have enough context info to intelligently (partially) wrap any trailing text anyways.

mickaelistria:
> Also it is called "enabling auto close brackets" but its not always a bracket right?

Should be just "Enable auto closing". A PR to fix the label would be welcome.

> But in the Java editor I am not so bothered by that, it seems they know a bit better when or when not to do it.

It knows *much* better when to do it or not. As it has access to an AST, it can evaluate the impact of auto-insert vs not auto-insert in AST, with something like `if (doesAutoInsertDetachNextTokenFromItsParentAndThusProbablyBreaksAST) simpleInsert(); else autoInsert();`. I don't think we have this here.

> do you think it makes sense we change it so brackets are only automatically closed if inserted at the end of a line of line? I don't think we have enough context info to intelligently (partially) enclose any trailing text anyways.

Even that is both a bit weak and error-prone IMO, I don't think it's noticeably better than the current state. But so far I also don't have any idea of what would be a better strategy.

sebthom:
[...]

PR #415, opened 2022-06-13
Only trigger auto-closing at end of line or before whitespace char

sebthom:
Addresses #305

Example:
Animation


mickaelistria:
good idea.

PR #417, opened 2022-06-13
Move all classes of languageconfiguration plugin to internal package

sebthom:


sebthom:
@mickaelistria I could not find references to these classes in any other eclipse plugins (also checked wild web developer)

PR #421, opened 2022-06-14
Improve toggle comment menu contributions

sebthom:
1) Currently the Add/Remove Block Comment / Toggle Line Comment commands are only accessible via their registered short cuts and not visible in the Generic Editors popup menu despite there registration in the plugin.xml for that exactly that purpose. 
	I guess because of a bug in Eclipse the menu items are not displayed. To fix this I had to specify the `label` attribute for the `<command>` element under the `<menuContribution>`. Apparently the label value from the command's definition is ignored.

	  ```xml
	  <menuContribution allPopups="true" locationURI="popup:org.eclipse.ui.genericeditor.source.menu?after=additions">
	     <!-- because of an eclipse bug, the "label" attribute needs to be explicitly set, otherwise entries don't show up in popup menu -->
	     <command commandId="org.eclipse.tm4e.languageconfiguration.toggleLineCommentCommand"
	        label="%LanguageConfiguration.toggleLineComment.name">
	  ```

	image

2) I also added menuContributions that show a `Source` menu in the main window. Interestingly for these contributions the `label` did not need to be specified again and Eclipse simply used the label from the command definition.

	image

3) I reordered the entries of the plugin.xml and added some comments for better comprehensibility


mickaelistria:
Thanks!

PR #427, opened 2022-06-30
Update Jenkinsfile

sebthom:
I updated the Jenkinsfile similar to the one found in the LSP4E project.

1. I removed the deprecated/unused option `-Dmaven.test.error.ignore=true`
2. added build timeout `timeout(time: 20, unit: 'MINUTES')`
3. Use a DOWNLOAD_AREA variable in the deploy stage to avoid unnecessary repetition of the absolute path


mickaelistria (changes requested):
Version bumps without changes are to be avoided.

sebthom:
@mickaelistria I removed the version bumps, now the build fails again with ` Only qualifier changed for (org.eclipse.tm4e.core/0.5.0.202207051504). Expected to have bigger x.y.z than what is available in baseline (0.5.0.202206232130)`

mickaelistria:
The GitHub actions needs to be configured to checkout proper commit and not do a shallow clone with a dummy mock commit as it breaks version computation.
Jenkins is more trustworthy (that's also one reason I'm not convinced that adding GitHub action brings more value than trouble and confusion).

PR #439, opened 2022-09-07
Fix indentation problem with onEnterRule

angelozerr:
Fix indentation problem with onEnterRule

Signed-off-by: azerr <azerr@redhat.com>


angelozerr:
I did that very quickly and it requires some write of tests, but it should fix https://github.com/eclipse/wildwebdeveloper/issues/599

See the following demo:

OnEnterRuleWithTSComments

mickaelistria:
I would be fine merging it even without tests if it was tested manually.

angelozerr:
> I would be fine merging it even without tests if it was tested manually.

If you think the behavior is good, feel free to merge it, but I cannot promise you that it will provide some troubles. We need to write tests (and perhaps there are some vscode tests that we should translate for tm4e).

PR #440, opened 2022-09-08
Use the inner parent as SWT parent in UI language configuration preference

angelozerr:
Use the inner parent as SWT parent in UI language configuration preference

Signed-off-by: azerr <azerr@redhat.com>


angelozerr:
Before the PR:

image

After the PR:

image

Issue #445, opened 2022-10-03
Toggle comment doesn't toggle back off comments but keeps adding comments chars


mickaelistria:
Using Wild Web Developer and an XML file such as
```xml
<a>
  <b/>
  <c/>
</a>
```
I put my cursor on line 2 and 3 and hit Ctrl+Shift+C, I get what's expected:
```xml
<a>
<!--  <b/>-->
<!--  <c/>-->
</a>
```
Now I do it again and expect to get back
```xml
<a>
  <b/>
  <c/>
</a>
```
instead I get
```xml
<a>
<!--<!--  <b/>-->-->
<!--<!--  <c/>-->-->
</a>
```

PR #447, opened 2022-10-05
Toggle comment doesn't toggle back off comments but keeps adding comm…

vrubezhny:
…ents chars #445

Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>

### DIFF
--- a/org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/utils/TextUtils.java
+++ b/org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/utils/TextUtils.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2015-2017 Angelo ZERR.
+ * Copyright (c) 2015-2022 Angelo ZERR and others.
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
  * which is available at https://www.eclipse.org/legal/epl-2.0/
@@ -182,4 +182,28 @@ public final class TextUtils {
 		return new TabSpacesInfo(4, false);
 	}
 
+	/**
+	 * Determines if all the characters at any offset of the specified document range are space or tab characters.
+	 *
+	 * @param document the document to search in
+	 * @param line zero-based document line number
+	 *
+	 * @return  <code>true</code>  if all the characters of the specified document range are the whitespace 
+	 * 				characters, otherwise returns <code>false</code>
+	 */
+	public static boolean isBlankLine(final IDocument document, int line) {
+		try {
+			int offset = document.getLineOffset(line);
+			final int lineEnd = offset + document.getLineLength(line);
+			while (offset < lineEnd) {
+				if (!Character.isWhitespace(document.getChar(offset))) {
+					return false;
+				}
+				offset++;
+			}
+		} catch (BadLocationException e) {
+			// Ignore,  forcing a positive result
+		}
+		return true;
+	}
 }
--- a/org.eclipse.tm4e.feature/feature.xml
+++ b/org.eclipse.tm4e.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.tm4e.feature"
       label="%featureName"
-      version="0.5.1.qualifier"
+      version="0.5.2.qualifier"
       provider-name="%featureProvider"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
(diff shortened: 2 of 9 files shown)


org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/utils/TextUtils.java
```diff
@@ -182,4 +182,28 @@ public static TabSpacesInfo getTabSpaces(@Nullable final ITextViewer viewer) {
 		return new TabSpacesInfo(4, false);
 	}
 
-}
+	/**
+	 * Determines if all the characters at any offset of he specified document range are space or tab characters.
```
sebthom, line 186: minor typo "of he specified" -> "of the specified"

org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/utils/TextUtils.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2015-2017 Angelo ZERR.
+ * Copyright (c) 2015-2022 Angelo ZERR.
```
sebthom, line 2: shouldn't this actually say "and others."?

org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/utils/TextUtils.java
vrubezhny: Thanks. I'll fix it.

org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/utils/TextUtils.java
vrubezhny: I have nothing against adding "and others" text, but IMHO it's something that author is to decide. If @angelozerr wouldn't mind - I can change.

sebthom:
@mickaelistria PR looks good to me. can you verify that it solves the issue for you?

mickaelistria:
It does fix the issue I currently face, but I noticed 2 pitfalls (that we may cover in other PRs if that's preferred):
* The selection is messed up after the toggle comment action. To reproduce select 2 lines in an XML file, use Ctrl+Shift+C and selection moves only to a subset of the 2nd line.
* Toggle is applied line by line, so if one line has comment and other hasn't and both are selected and then toggle comment is invoked, then 1 line is un-commented and the other is commented. I believe the usual expectation, as built by JDT is that the same operation (comment or uncomment according to eg 1st line) is applied to all selected lines.

vrubezhny:
> * Toggle is applied line by line, so if one line has comment and other hasn't and both are selected and then toggle comment is invoked, then 1 line is un-commented and the other is commented. I believe the usual expectation, as built by JDT is that the same operation (comment or uncomment according to eg 1st line) is applied to all selected lines.

.. and this is exactly the same algorithm was applied previously on Toggle Single-Line comment before I changed the behavior. 
If in Java source we have selected several lines and some of them are already starting with a single-line comment a new single-line comment is added to such a line - comments are duplicating... which was a problem that we were solving in TM4E...

The only difference I see (from Java case) is that for xml-based text we don't have a single-line comment template - there is only a multi-line comments, while in Java we have both - single-line and multi-line comments.

So a solution I see is:
- First row in text selection defines the operation to be performed - add or remove comments - for the rest of the selected rows
- if a single-line comment exists in TM4E for a document type - perform the Toggle single-line comments action exactly like it's done in Java (including the single-line comment duplication)
- if NO single-line comment exists in TM4E for a document type - then use multi-line comment template and NOT duplicate the comment for a raw that is already commented out.

WDYT?

mickaelistria:
This looks good to me. Any reason why it's still a draft? Anything left to do before we can merge?

vrubezhny:
> This looks good to me. Any reason why it's still a draft? Anything left to do before we can merge?

The selection is still messed up if a line has a block comment inside the text or more than one block comment. So I'm working on make it  to be similar to what we have for Java editor (regarding the block comments) including the selection change.

mickaelistria:
Couldn't we merge this increment, that I believe does cover very well the vast majority of cases, and track that other issue in a separate ticket/PR?

vrubezhny:
I think yes, this PR can be merged.
Other changes will follow in a new PR

mickaelistria:
Thanks!

PR #451, opened 2022-10-08
Fix syntax colouring when inline comments contain multi-byte characters + numerical constants/punctuation

shapayev:
When you type something like the following code in C/C++:

```
struct S {
    //! мышка (just a mouse, only 1)
    char mouse; // syntax colouring of this string will be broken
    //! кошка (just a cat, only 1)
    char cat; // syntax colouring of this string will be broken
};
```

the syntax colouring of a line below the triggering string will be broken. I've managed to reproduce this behaviour with the latest build of Eclipse Corrosion and the most current TM4E snapshot, you can see it in the animated gif below.

cat_and_mouse

EDIT: Double-checked this today with the provided example: in my case (Eclipse CDT with a small patch to connect the extension point of the generic editor to the TMPresentationReconciler of TM4E) without this fix I also had the following exception:
```
java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4899)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.custom.StyledText.setStyleRanges(StyledText.java:10665)
	at org.eclipse.swt.custom.StyledText.replaceStyleRanges(StyledText.java:8238)
	at org.eclipse.jface.text.TextViewer.addPresentation(TextViewer.java:4777)
	at org.eclipse.jface.text.TextViewer.changeTextPresentation(TextViewer.java:4854)
	at org.eclipse.tm4e.ui.text.TMPresentationReconciler.applyTextRegionCollection(TMPresentationReconciler.java:692)
	at org.eclipse.tm4e.ui.text.TMPresentationReconciler.colorize(TMPresentationReconciler.java:596)
	at org.eclipse.tm4e.ui.text.TMPresentationReconciler$InternalListener.colorize(TMPresentationReconciler.java:410)
```
And so debugging this one had lead me to the LineTokenizer trying to handle the captured token of negative length.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,6 +90,11 @@
 				</executions>
 				<dependencies>
 					<dependency>
+						<groupId>org.apache.maven.surefire</groupId>
+						<artifactId>surefire-junit-platform</artifactId>
+						<version>3.0.0-M7</version>
+					</dependency>
+					<dependency>
 						<groupId>org.junit.jupiter</groupId>
 						<artifactId>junit-jupiter-engine</artifactId>
 						<version>5.9.1</version>
--- a/org.eclipse.tm4e.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.tm4e.core/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-SymbolicName: org.eclipse.tm4e.core
-Bundle-Version: 0.5.0.qualifier
+Bundle-Version: 0.5.1.qualifier
 Require-Bundle: org.apache.batik.css;bundle-version="1.9.1";resolution:=optional,
  org.apache.batik.util;bundle-version="1.9.1";resolution:=optional,
  com.google.gson;bundle-version="2.9.0",
--- a/org.eclipse.tm4e.core/pom.xml
+++ b/org.eclipse.tm4e.core/pom.xml
@@ -10,7 +10,7 @@
 
 	<artifactId>org.eclipse.tm4e.core</artifactId>
 	<packaging>eclipse-plugin</packaging>
-	<version>0.5.0-SNAPSHOT</version>
+	<version>0.5.1-SNAPSHOT</version>
 
 	<profiles>
 		<profile>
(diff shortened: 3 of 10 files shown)


sebthom:
Hi Alexander,
thank you for your contribution. Can you please describe the issue and provide some examples (e.g. screenshots) of the issue you are seeing?

shapayev:
Hi Sebastian,
Sorry for the lack of initial description. Should I open a separate issue for the bug that this pull request proposes to fix?

In brief, when you type something like this code in C/C++:
// ---------- CODE BEGINS ----------
char cat; // кошка (just a cat, only 1)
char mouse; // syntax colouring of this string will be broken
// ---------- CODE ENDS ----------
you'll have a SWT Exception extending from the LineTokenizer due to trying to handle the captured token of negative length. Also the syntax colouring of one or few lines below the triggering string will be broken.

I shall double check this particular example tomorrow (and shall attach screenshots if needed) as I don't have an access to the Eclipse IDE with the current TM4E right now. The example string with this particular multi-byte comment may or may not trigger the bug depending on the number of multi-byte characters and their relative position to the other symbols (numbers and/or punctuation) attempted to be captured as separate tokens.

sebthom:
We don't need a separate issue for this. Comments in the PR are fine.

mickaelistria:
Do you think you could add some tests that cover this bug+fix? I'm afraid it's too easy to introduce a regression later without a test.

pom.xml
```diff
@@ -89,6 +89,11 @@
 					</execution>
 				</executions>
 				<dependencies>
+					<dependency>
+						<groupId>org.apache.maven.surefire</groupId>
```
sebthom, line 93: Out of curiosity, why did you need to add this dependency?

sebthom:
The PR looks good to me. Thanks for adding the test cases. 

The internal tests were not skipped by intention. Thanks for the fix.

shapayev:
Sorry, can't comment inline at the moment so shall try to explain it here.
I shall fix the comment, the returned value looks good for me ;)
The dependency was added to force using JUnit5 by surefire. Could not deduce why the auto-detection is not working otherwise. Maybe there is a better fix, this one has turned to be the most compact.

sebthom:
I removed my inline comments again. All good. Thanks again for the PR. Identifying this issue and fixing it was very good work!

shapayev:
Thank you much very much for your warm reception! I am very glad I could be of any help to you.

PR #452, opened 2022-10-10
fix scopeName of Regular Expressions (JavaScript)

shapayev:
This fix is the tiniest one. When trying to open any *.txt file in generic editor with the org.eclipse.tm4e.language_pack installed one can get the following error:
fail_screen


sebthom:
Good catch! Please also bump the version of the feature.

sebthom:
Thank you!!

PR #459, opened 2022-10-27
Fix Text Selection after toggling line comment when only block commen…

vrubezhny:
…t defined

The Text Selection is preserved as much as possible after the toggle line comment action when language configuration defines only block comment (so it's used instead of line comments)

Also fixes: https://github.com/eclipse/wildwebdeveloper/issues/909

### DIFF
--- a/org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/ToggleLineCommentHandler.java
+++ b/org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/ToggleLineCommentHandler.java
@@ -105,137 +105,154 @@ public class ToggleLineCommentHandler extends AbstractHandler {
 			try {
 				switch (command.getId()) {
 				case TOGGLE_LINE_COMMENT_COMMAND_ID: {
-					final var lineComment = commentSupport.getLineComment();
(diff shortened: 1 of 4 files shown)


org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/ToggleLineCommentHandler.java
```diff
@@ -250,6 +354,40 @@ private boolean areLinesCommented(final IDocument document, final ITextSelection
 		return true;
 	}
 
+	private Set<ITypedRegion> getBlockCommenPartssForLine(final IDocument document, final int line,
```
sebthom, line 357: one ` s` too much after Parts?

sebthom:
LGTM, however I would like to wait for @mickaelistria feedback before merging

org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/ToggleLineCommentHandler.java
vrubezhny: Sorry. Getting blind. 
Fixed.

PR #466, opened 2022-11-10
Add public API to get default theme

rubenporras:
Would it be possible to add this API to help building a semantic reconciler in LSP4E (see https://github.com/eclipse/lsp4e/pull/253)


rubenporras:
I would actually an API for getting the theme applied to a particular editor/document, and not just the default one. Any help on exposing such API? Or can theme already be retrieved with the existing APIs? I could not find the way of doing it but may someone with more expertise knows?

Thanks

rubenporras:
yes, you are right, that should work. Any idea on how get the theme which is actually used by an editor and not just the default one?

mickaelistria:
> ny idea on how get the theme which is actually used by an editor and not just the default one?

I don't think it's easy; the only current path to get that would be to get the underlying viewer for the editor, and then to use reflection in order to access its `fPresentationReconciler` field, and then check for this field whether it's a TMPresentationReconciler and then call `getTokenProvider` and verify whether it can be cast to ITheme`.

Issue #478, opened 2022-12-18
Updated versions available for the target platform


merks:
The following minor version updates are available for the target platform:

 - [com.google.code.gson](https://repo1.maven.org/maven2/com/google/code/gson/)
    - [gson](https://repo1.maven.org/maven2/com/google/code/gson/gson/) **[2.9.1](https://repo1.maven.org/maven2/com/google/code/gson/gson/2.9.1)** < [2.10](https://repo1.maven.org/maven2/com/google/code/gson/gson/2.10/)
 - [org.jruby.joni](https://repo1.maven.org/maven2/org/jruby/joni/)
    - [joni](https://repo1.maven.org/maven2/org/jruby/joni/joni/) **[2.1.43](https://repo1.maven.org/maven2/org/jruby/joni/joni/2.1.43)** < [2.1.44](https://repo1.maven.org/maven2/org/jruby/joni/joni/2.1.44/

PR #479, opened 2022-12-18
Update the dependencies to newer minor versions

merks:
Fixes https://github.com/eclipse/tm4e/issues/478


akurtakov:
Thanks @merks !

PR #481, opened 2023-01-07
Contribute "Source" menu entry in ui plugin instead of langcfg plugin

sebthom:
This allows other plugins to contribute to the menu without depending on the presence of the languageconfiguration plugin


mickaelistria:
Just for the record, I believe the initial goal was to keep Syntax Highlighting and Language Configuration separated so one could use or or the other without need to have both installed.
In practice, I'm not sure we need to care much about that case these days, so I'm OK with this change.

sebthom:
@mickaelistria the reason why I wanted to move the entry to the ui plugin is, that I am working on contributing a "Format" entry to that menu from the lsp4e plugin. the lsp4e plugin has currently no dependency to the tm4e.langcfg plugin. So I can either add that dependency or move the "Source" menu entry definition to the tm4e ui plugin. which way do you prefer?

PR #500, opened 2023-03-07
fix: Increase "depth" of content type to lower its priority

### LINKED ISSUE #499: TM4E language_pack definitions taking priority over CDT ones in some cases
If you open a `.hpp` file in Eclipse with both CDT and TM4E installed it opens in the generic editor instead of the CEditor.

My reading of the TM4E code is that the intention was that the language_pack definitions were supposed to be lower [priority](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/extension-points/org_eclipse_core_contenttype_contentTypes.html) that things the user already had installed.

However, it turns out that the way Eclipse chooses a content type depends on the ["depth"](https://github.com/eclipse-platform/eclipse.platform/blob/9839c18f52687db6f949bdb56a59a5d20cd7c938/runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentTypeCatalog.java#L43-L60) of the content type by considering how many base types it has. And CDT's content type for `.hpp` files is deeper than TM4E's one.

Therefore, in the absence of touching [ContentTypeCatalog](https://github.com/eclipse-platform/eclipse.platform/blob/9839c18f52687db6f949bdb56a59a5d20cd7c938/runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentTypeCatalog.java#L43-L60) code that hasn't been changed since 2005, I will provide a PR to resolve the issue by creating artificial depth.

jonahgraham:
Fixes #499

### DIFF
--- a/org.eclipse.tm4e.language_pack.feature/feature.xml
+++ b/org.eclipse.tm4e.language_pack.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.tm4e.language_pack.feature"
       label="%featureName"
-      version="0.1.1.qualifier"
+      version="0.1.2.qualifier"
       provider-name="%featureProvider"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/org.eclipse.tm4e.language_pack.feature/pom.xml
+++ b/org.eclipse.tm4e.language_pack.feature/pom.xml
@@ -10,5 +10,5 @@
 
 	<artifactId>org.eclipse.tm4e.language_pack.feature</artifactId>
 	<packaging>eclipse-feature</packaging>
-	<version>0.1.1-SNAPSHOT</version>
+	<version>0.1.2-SNAPSHOT</version>
 </project>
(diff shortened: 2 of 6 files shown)


jonahgraham:
I ran the pom.xml in the update directory, but there were lots of other changes that were unrelated, so I only included the relevant plugin.xml changes.

mickaelistria:
I think the issue lies more in Platform: the priority should be taken into account when returning content-type.

jonahgraham:
> I think the issue lies more in Platform: the priority should be taken into account when returning content-type.

The priority is taken into account, just it is secondary in the sort to depth. 

Do you think there is any chance of consensus in Platform to change the algorithm on resolving content type selection after nearly 2 decades? I am not sure I want to clean up the fall out from the unintended consequences of that.

mickaelistria:
> Do you think there is any chance of consensus in Platform to change the algorithm on resolving content type selection after nearly 2 decades? 

I think if you look at history, there were some changes recently without strong debate. I think the current behavior qualifies as a bug as per specification of the extension point. Depth should be secondary to priority when content-types are from different branches.
I remember some other people have fairly complained about the depth being more important than priority in some cases; I didn't find a ticket, but I think it's easy to reach a technical consensus.
I think it's worth submitting a PR anyway, and if some other committer approves it, it would be merged.

jonahgraham:
With the fix for https://github.com/eclipse-platform/eclipse.platform/issues/151 merged this is now resolved in platform.

PR #502, opened 2023-03-08
fix: Regenerate language_pack

### LINKED ISSUE #501: language_pack is outdated?
While working on #500 I had to regenerate the language_pack. There are lots of differences between the current git HEAD and what is generated in the update directory.

Is this supposed to be done on a regular basis? How do we validate such changes?

I will provide a PR that does the update, but it isn't validated.

jonahgraham:
This was done by running `mvn package` in `_update`.

The autogenerated files have one small error that was
previously fixed in https://github.com/jonahgraham/tm4e/commit/242dcf1265231f765c69dea7ecd495e6620cd4b2
and this change therefore preserves that fix by excluding this
diff from the commit:

```diff
diff --git a/org.eclipse.tm4e.language_pack/javascript/Regular Expressions (JavaScript).tmLanguage b/org.eclipse.tm4e.language_pack/javascript/Regular Expressions (JavaScript).tmLanguage
index 54c7651c..1dda7806 100644
--- a/org.eclipse.tm4e.language_pack/javascript/Regular Expressions (JavaScript).tmLanguage
+++ b/org.eclipse.tm4e.language_pack/javascript/Regular Expressions (JavaScript).tmLanguage
@@ -230,7 +230,7 @@
                </dict>
        </dict>
        <key>scopeName</key>
-       <string>lngpck.source.js.regexp</string>
+       <string>source.js.regexp</string>
        <key>uuid</key>
        <string>AC8679DE-3AC7-4056-84F9-69A7ADC29DDD</string>
 </dict>
```

And bumping version numbers due to updated language packs

Fixes https://github.com/eclipse/tm4e/issues/501


mickaelistria:
@howlger What do you think?

jonahgraham:
> +1, (based on the code review without testing): `org.eclipse.tm4e.language_pack/javascript/Regular Expressions (JavaScript).tmLanguage` is incorrect generated and overrides the manual fix [242dcf1](https://github.com/eclipse/tm4e/commit/242dcf1265231f765c69dea7ecd495e6620cd4b2). It has not been changed, so please exclude this file.

I have updated the commit and the description of the PR. An updated look over it would be appreciated.

> 
> +1 also for @sebthom's idea

+1 From me too. 

> 
> Independent from this, would it be better to have separate plugins and features, instead of a single plugin/feature containing everything?

If you mean having as many plug-ins as there are languages in the language pack, I'm not keen on that idea.

howlger:
> I have updated the commit and the description of the PR. An updated look over it would be appreciated.
>
Looks good to me. It would be good to test this by opening a simple example for each language/format so that the grammar is loaded. If I remember correctly, not all regular expressions work; _lookbehind_ of an unknown number of characters is not supported and causes an error when loading the grammar. But that might have changed in the meantime. @sebthom, please correct me if this is not true. If you'd like me to test it rather than just roughly look it over, I can do that the week after next.

sebthom:
@howlger I'd appreciate if you could test the changes. it would actually also be nice to have an automated test that tries to load/parse the grammars.

mickaelistria:
TM4E is switching to `main` as main branch instead of `master`. Master branch will be removed soon. Can you please retarget this PR towards `main` branch?

howlger:
+1

Sorry for the delay. I tested this pull request manually with sample files for the different formats.

I will try to fix the generation to avoid overwriting of the manual fix https://github.com/eclipse/tm4e/commit/242dcf1265231f765c69dea7ecd495e6620cd4b2, and if I am successful, I will submit it as a separate pull request. After that, I also plan to look at how to automatically test it, to at least find the regular expressions that cause errors at runtime. But I can't promise anything and would be happy if someone else would do that.

sebthom:
Superseded by https://github.com/eclipse/tm4e/pull/550/

PR #506, opened 2023-03-13
refact: add checkstyle and fix findings

sebthom:


mickaelistria:
I'm not very thrilled with the idea of mandating CheckStyle and requiring it as nature (so someone who doesn't have Checkstyle cannot contribute). Isn't there a way to configure checkstyle on client side without strongly requiring it for development?

mickaelistria:
Please be aware that for many users (including me), Checkstyle is an annoyance that reports tons of non-problems and fixes 0 issues; many people use other static analysis tools -such as SonarLint- with better results and wouldn't enjoy having to mess with CheckStyle rule.
You can look for example at the reaction of most Maven contributors when they contribute to it: they're pretty annoyed by enforced styling which is a burden for their contribution, some do give up because of it. I classify introduction of checkstyle as a step in that direction, a wrong one IMO.

sebthom:
Based on my 3 decades in software development and the millions of LoCs I wrote I wholeheartedly disagree. A consistent source code greatly improves comprehensibility. Also checkstyle despite it's name does not only check "style" but also provides rules that prevent actual errors.

I am fine with using another tool but would like to see some kind of automated minimal quality gate for PRs. This should also be in the interest of the Eclipse Foundation in respect the the upcoming legal changes in the EU https://blogs.eclipse.org/post/mike-milinkovich/product-liability-directive-more-bad-news-open-source

mickaelistria:
> I am fine with using another tool but would like to see some kind of automated minimal quality gate for PRs.

The tricky part is to agree on "minimal quality gate".
I don't mind some code analysis that blocks a merge if there is a probable bug in the code, but mandating CheckStyle in devtime (or even in build-time) seems to do much more than spotting bugs; and the extra things are a huge constraints to contributors which would reduce the time between they try to contribute and their PR is merged; and keeping that time as short as possible is a more important criterion of sustainability for the project than having final on parameters (that a compiler or a static analyzer can easily detect whether they're effectively final or not and just bug in case they are).
Note that it is possible to configure some problems severity on JDT, and as Tycho uses JDT compiler, it can fail the build for problems that are marked as errors.

PR #517, opened 2023-04-03
fix: On enter rule don't use the proper insert spaces / tab width when those settings are changed

### LINKED ISSUE #460: On enter rule don't use the proper insert spaces / tab width when those settings are changed
As the compute of insert spaces / tabs is cached here https://github.com/eclipse/tm4e/blob/cb5f4728cd3f0f1b2eaa4f2836519309ffacd20c/org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/LanguageConfigurationAutoEditStrategy.java#L283 there is a bug with on enter rule which is not updated when those settings are changed.

I noticed that while my editorconfig implementation. See the following demo:

JSONEditorConfigDemo

In the demo the jsonc file have spaces (we don't see that, but it was initialized with that). When this settings is changed with tabs and when we press on on enter between {}, it generates spaces instead of tabs. It is because https://github.com/eclipse/tm4e/blob/cb5f4728cd3f0f1b2eaa4f2836519309ffacd20c/org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/LanguageConfigurationAutoEditStrategy.java#L283  was cached.

To fix this issue we should avoid having a cache and just have the capability to read the tabs / insert spaces settings from the editor preferences store. This fixes requires https://github.com/eclipse-platform/eclipse.platform.text/pull/88

I remember too that we use uggly reflection to get information of insert spaces at https://github.com/eclipse/tm4e/blob/cb5f4728cd3f0f1b2eaa4f2836519309ffacd20c/org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/utils/TextUtils.java#L175 we should avoid doing that.

Another uggly code is here https://github.com/eclipse/tm4e/blob/cb5f4728cd3f0f1b2eaa4f2836519309ffacd20c/org.eclipse.tm4e.languageconfiguration/src/main/java/org/eclipse/tm4e/languageconfiguration/internal/LanguageConfigurationAutoEditStrategy.java#L293 to get the editor of the IAutoEditStrategy I created an issue at https://github.com/eclipse-platform/eclipse.platform.text/pull/108#issuecomment-1293957398 to get easily the editor just by implementing ITextEditorAware.

angelozerr:
fix: On enter rule don't use the proper insert spaces / tab width when those settings are changed

Fixes #460

Signed-off-by: azerr azerr@redhat.com


mickaelistria:
This seems to require a version bump.

angelozerr:
@sebthom if I understand I need to update only the languageconfiguration plugin to 0.5.4.qualifier is that?

sebthom:
Yes, you should only need to bump the version of the plugins you are touching.

angelozerr:
> Yes, you should only need to bump the version of the plugins you are touching.

Thanks @sebthom for your clarification. 

@mickaelistria I pushed a new commit with bump,let's see if CI build will work now.

angelozerr:
@mickaelistria @sebthom it should be good now.

mickaelistria:
Thank you!

PR #519, opened 2023-04-21
fix: Do not create error markers for comments

rubenporras:
Otherwise adding a comment to a file like
  // Need to check abc due to BUG 000233
or 
  // This unit tests makes sure BUG 000233 does not happen again

will add error markers to a resource. We have comments like this in our code base, and I do not think the users should see error markers because of them (indeed I think the content of a comment should never create an error marker)

### DIFF
--- a/org.eclipse.tm4e.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.tm4e.ui/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-SymbolicName: org.eclipse.tm4e.ui;singleton:=true
-Bundle-Version: 0.6.2.qualifier
+Bundle-Version: 0.6.3.qualifier
 Require-Bundle: org.eclipse.tm4e.core;bundle-version="0.5.2",
  org.eclipse.jface.text,
  org.eclipse.core.runtime,
--- a/org.eclipse.tm4e.ui/pom.xml
+++ b/org.eclipse.tm4e.ui/pom.xml
@@ -10,5 +10,5 @@
 
 	<artifactId>org.eclipse.tm4e.ui</artifactId>
 	<packaging>eclipse-plugin</packaging>
-	<version>0.6.2-SNAPSHOT</version>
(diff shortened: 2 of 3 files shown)


rubenporras:
@sebthom , could you review this PR? Thanks

sebthom:
please bump the plugin version

rubenporras:
I have bumped the plugin version

rubenporras:
All that is visible in the log is

Error: 8,209 [ERROR] Dependency license check failed. Some dependencies need to be vetted.
...
Error: 8,217 [ERROR] Failed to execute goal org.eclipse.dash:license-tool-plugin:0.0.1-SNAPSHOT:license-check (default-cli) on project org.eclipse.tm4e: Some dependencies must be vetted. -> [Help 1]

I have never seen this check and setting up local build will take some time for me. My change can for sure not change licenses or dependencies, so I  was hoping I would not need to do that. Would be possible to merge it even though the check fails, and that some TM4E maintainer looks at it when he has time?

Regards

sebthom:
@mickaelistria do you have a chance to check why the license check suddenly fails? I am a.f.k. at the moment.

sebthom:
Marking `BUG` as errors is intentional. I am of a different opinion regarding your statement *"I think the content of a comment should never create an **error marker**"*. From my perspective, the `BUG` keyword in the comment serves as a manual **error marker**, indicating that an error has been manually detected. Therefore, the IDE should interpret it as an error, just like an error found by a compiler or a static code analysis tool.

The issue in this particular case is that the `BUG` keyword is not used as a `BUG` marker, but rather in the description of a workaround to reference a reported bug in an external ticket system.

To address this, I see the following options:
1. You could change `BUG 000233` to `bug 000233` or `Bug 000233`  or `BUG000233` in your code base.
1. The code that parses the comment for markers could be modified so that it either:
    1. ignores words that match markers if they are not at the beginning of a comment line, or
    1. requires that markers are followed by a colon, e.g. `BUG: This method returns wrong values when called at night`
1. The currently static marker configuration could be made configurable per workspace/project via the UI. However, I am not willing to put the necessary effort into this one. PRs welcome :-)

Any thoughts?

sebthom:
superseded by #528

PR #522, opened 2023-04-27
fix: minor bugfix to ToggleLineCommentHandler.

andrewL-avlq:
Fix removeLineComments() to ensure that lines are processed in order.


mickaelistria:
thank you!

PR #528, opened 2023-05-05
feat: add Task Tag UI preferences

sebthom:
This PR adds the ability to add,remove,configure task tags per workspace via preferences:

image

### DIFF
--- a/.github/workflows/licensecheck.yml
+++ b/.github/workflows/licensecheck.yml
@@ -23,16 +23,18 @@ env:
 jobs:
 
   build:
+
     runs-on: ubuntu-latest
+
     steps:
     - name: Git Checkout
       uses: actions/checkout@v3 #https://github.com/actions/checkout
 
-    - name: Set up JDK 17
+    - name: "Install: JDK ${{ env.JAVA_VERSION }}"
       uses: actions/setup-java@v3 # https://github.com/actions/setup-java
       with:
-        distribution: 'temurin'
-        java-version: 17
+        distribution: temurin
+        java-version: ${{ env.JAVA_VERSION }}
 
     - name: "Cache: Local Maven Repository"
       uses: actions/cache@v3
@@ -41,8 +43,6 @@ jobs:
           ~/.m2/repository
           !~/.m2/**/*SNAPSHOT*
         key: ${{ runner.os }}-mvnrepo-${{ hashFiles('**/pom.xml') }}-${{ hashFiles('**/target-platform/tm4e-target.target') }}
-        restore-keys: |
-          ${{ runner.os }}-mvnrepo-
 
     - name: "Install: Maven"
       uses: stCarolas/setup-maven@v4.5 # https://github.com/stCarolas/setup-maven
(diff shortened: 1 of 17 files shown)


sebthom:
@rubenporras can you please give this a try

.github/workflows/licensecheck.yml
```diff
@@ -53,11 +53,10 @@ jobs:
       run: |
         set -eu
 
-        MAVEN_OPTS="${MAVEN_OPTS:-}"
-        MAVEN_OPTS="$MAVEN_OPTS -XX:+TieredCompilation -XX:TieredStopAtLevel=1" # https://zeroturnaround.com/rebellabs/your-maven-build-is-slow-speed-it-up/
-        MAVEN_OPTS="$MAVEN_OPTS -Djava.security.egd=file:/dev/./urandom" # https://stackoverflow.com/questions/58991966/what-java-security-egd-option-is-for/59097932#59097932
-        MAVEN_OPTS="$MAVEN_OPTS -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss,SSS" # https://stackoverflow.com/questions/5120470/how-to-time-the-different-stages-of-maven-execution/49494561#49494561
-        MAVEN_OPTS="$MAVEN_OPTS -Xmx1024m -Djava.awt.headless=true -Djava.net.preferIPv4Stack=true -Dhttps.protocols=TLSv1.2"
+        MAVEN_OPTS+=" -XX:+TieredCompilation -XX:TieredStopAtLevel=1" # https://zeroturnaround.com/rebellabs/your-maven-build-is-slow-speed-it-up/
```
rubenporras, line 56: TieredCompilation is the default since java 8. I think +TieredCompilation can be removed for sure. I think the stop at level 1 as well.

rubenporras:
@sebthom , thanks for doing the UI config backed by preferences. If you do not mind, I would do some testing after it is merged using the snapshot that will be built afterwards, that is a bit easier for me.

rubenporras:
I still have some trouble understanding the differentiation between Task and Problem. I think that a single category (Task) with three priorities is simpler and more understandable, and also enough, as in the JDT:

image

I would also change the text of the dialog for one more descriptive, like the one in the screenshot.

I also tested the JDT again, and even for prio high the icon is not the one of an error marker, which is what I would expect as well.

sebthom:
@rubenporras I changed the text in the dialog.

Problem markers appear in the problems view and are annotated on the icons in the project explorer.

rubenporras:
Thanks. I understand that now. It still feels to me that having just task markers is the best thing, and reserve problem makers for the language server (be it compiler, code analyzer, etc.). 

The second best option to me would be to have the default to just use tasks and leave problems for power users that really want to have that.

This is based on my personal preference, but also in the experience of using JDT and other tools, so far I do not know about any that show error markers for comments.

rubenporras:
I tested today on the released TM4E, and it does not seem to work, I have comments such us (we use -- as single line comment separator)
     -- BUG 123
     -- TODO this and that
and I can see that the matcher org.eclipse.tm4e.ui.internal.utils.MarkerUtils.updateTextMarkers(TMDocumentModel, int).matcher on line 135 does not find anything when commentText contains"BUT 123" or  "TODO this and that". 
The MARKERCONFIG_BY_TAG contains the default tags (BUG, TODO...) so I am not sure what is happening.

PR #532, opened 2023-05-15
fix: task tags are ignored if not at the start of a comment

sebthom:
as of the last release task tags are only evaluated if they are at the beginning of a comment.
this PR restores the previous behaviour of evaluating task tags anywhere in the comment.

### DIFF
--- a/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/utils/MarkerUtils.java
+++ b/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/utils/MarkerUtils.java
@@ -50,7 +50,7 @@ public final class MarkerUtils {
 				MARKERCONFIG_BY_TAG.put(markerConfig.tag, markerConfig);
 			}
 			TAG_SELECTOR_PATTERN = Pattern
-					.compile("^\\s*(" + MARKERCONFIG_BY_TAG.keySet().stream().collect(Collectors.joining("|")) + ")\\b");
+					.compile("\\b(" + MARKERCONFIG_BY_TAG.keySet().stream().collect(Collectors.joining("|")) + ")\\b");
 		}
 	}
 
--- a/org.eclipse.tm4e.feature/feature.xml
+++ b/org.eclipse.tm4e.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.tm4e.feature"
       label="%featureName"
-      version="0.5.4.qualifier"
+      version="0.5.5.qualifier"
       provider-name="%featureProvider"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/org.eclipse.tm4e.feature/pom.xml
+++ b/org.eclipse.tm4e.feature/pom.xml
@@ -10,5 +10,5 @@
 
 	<artifactId>org.eclipse.tm4e.feature</artifactId>
 	<packaging>eclipse-feature</packaging>
-	<version>0.5.4-SNAPSHOT</version>
+	<version>0.5.5-SNAPSHOT</version>
 </project>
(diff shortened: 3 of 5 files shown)


org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/utils/MarkerUtils.java
```diff
@@ -50,7 +50,7 @@ public static void reloadMarkerConfigs() {
 				MARKERCONFIG_BY_TAG.put(markerConfig.tag, markerConfig);
 			}
 			TAG_SELECTOR_PATTERN = Pattern
-					.compile("^\\s*(" + MARKERCONFIG_BY_TAG.keySet().stream().collect(Collectors.joining("|")) + ")\\b");
+					.compile("\\b*(" + MARKERCONFIG_BY_TAG.keySet().stream().collect(Collectors.joining("|")) + ")\\b");
```
rubenporras, line 53: Hi @sebthom ,

If we remove the "^", I think we not need the "*", I think \b should be sufficient, instead of \b*.

I have a bit more time this morning, and I think that the problem is not the regex, but the fact that the comment delimiter is part of the comment string, `commentText` in my case contains "-- TODO this and that", and not " TODO this and that". So "^\\s" will never match any comment.

Is that maybe the problem? A bit above there is a check

```
					if (commentText.length() < 3)
						continue;
```

I am not sure how it all fits together but if the comment separator is going to be part of the string, that condition will almost always be true, so one could as well remove it and rely on the matcher, I do not think performance wise it is much of a speedup, since in this case most comments will pass that condition. In addition there is no hint that task patterns must have more than n characters.


Sorry for the bunch of unstructured information/questions. I do not know the code at all, so I am just commenting just in case you have overlooked something. If not, you can ignore my comments.

rubenporras:
BTW: Our single line comment is defined as
```
        "linecomment": {
            "name": "comment.line.double-dash.avqscript",
             "begin": "(^[ \\t]+)?(?=--)",
             "beginCaptures": {
               "1": {
                 "name": "punctuation.whitespace.comment.leading.ts"
               }
             },
            "end": "(?=$)"
        },
```
Then if the line has the text "     -- TODO this and that" then the variable `commentText` has first the content "     ", and the token is "comment.line.double-dash.avqscript-punctuation.whitespace.comment.leading.ts" and then "-- TODO this and that" nd the token is "comment.line.double-dash.avqscript".

rubenporras (approved):
This looks good to me now. IIRC, it is the same regex that we had before the last release

org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/utils/MarkerUtils.java
sebthom: > If we remove the "^", I think we not need the "*", I think `\b` should be sufficient, instead of `\b*`.

That is correct. I changed it. Thanks!

> I have a bit more time this morning, and I think that the problem is not the regex, but the fact that the comment delimiter is part of the comment string, `commentText` in my case contains "-- TODO this and that", and not " TODO this and that". So `^\\s*` will never match any comment.

Interesting. In the languages that I am using/testing (Dart and Haxe), the `commentText` does not contain the comment delimiter. Maybe it depends on how the textmate rule is defined.

> Is that maybe the problem? A bit above there is a check
> 
> ```
> 					if (commentText.length() < 3)
> 						continue;
> ```
> 
> I am not sure how it all fits together but if the comment separator is going to be part of the string, that condition will almost always be true, so one could as well remove it and rely on the matcher, I do not think performance wise it is much of a speedup, since in this case most comments will pass that condition. In addition there is no hint that task patterns must have more than n characters.

I am hitting that condition quite often in Dart and Haxe so I would prefer to keep that check.

sebthom:
> This looks good to me now. IIRC, it is the same regex that we had before the last release
That was actually my intention with this PR but I apparently needed some attempts...

PR #535, opened 2023-05-17
fix: Pass regex warnings to Logger instead of printing it to stderr

### LINKED ISSUE #534: Regular expression warnings are printed to stderr
If a language grammar contains regular expressions which lead to warnings in `org.joni.Regex', they are always printed to stderr.

This can result in spamming the console, e.g. https://github.com/eclipse-cdt/cdt-lsp/issues/88.

I suggest to provide a custom `WarnCallback` which passes the warnings to a `java.lang.System.Logger` with level `Level.WARNING`.

I will provide a PR soon.

ddscharfe:
Fixes #534


ddscharfe:
I forgot to increase the bundle version.

Issue #540, opened 2023-06-07
Coloring exhibits quadratic behavior and blocks the SWT thread


szarnekow:
Opening a medium[1] sized JSON file in Eclipse might block the UI seemingly indefinitely due to `AbstractModelLines.getOrNull` which attempts query a linked list by index.

Relevant part of the stack:
```
"main" #1 [259] prio=6 os_prio=31 cpu=255346.03ms elapsed=871.90s tid=0x00007fadbe021400 nid=259 runnable  [0x0000000205a14000]
   java.lang.Thread.State: RUNNABLE
	at java.util.LinkedList.node(java.base@20.0.1/LinkedList.java:575)
	at java.util.LinkedList.get(java.base@20.0.1/LinkedList.java:481)
	at org.eclipse.tm4e.core.model.AbstractModelLines.getOrNull(AbstractModelLines.java:149)
	- locked <0x0000000769f00080> (a java.util.LinkedList)
	at org.eclipse.tm4e.core.model.TMModel.getLineTokens(TMModel.java:284)
	at org.eclipse.tm4e.ui.text.TMPresentationReconciler.colorize(TMPresentationReconciler.java:540)
	at org.eclipse.tm4e.ui.text.TMPresentationReconciler$InternalListener.textChanged(TMPresentationReconciler.java:336)
	at org.eclipse.jface.text.TextViewer.updateTextListeners(TextViewer.java:2785)
	at org.eclipse.jface.text.TextViewer.invalidateTextPresentation(TextViewer.java:3405)
	at org.eclipse.jface.text.TextViewer.initializeWidgetContents(TextViewer.java:3453)
	at org.eclipse.jface.text.TextViewer.setVisibleDocument(TextViewer.java:3492)

```

[1] To quantify this: 800k lines, 22MB; we might call it a big JSON file.

It shall be noted that a quick glance over AbstractModelLines.java indicates that almost all of the usages of the internal linked list are index based. It might warrant a refactoring of the entire class to avoid performance penalty that is imposed on all methods with the current data structure.

mickaelistria:
What can be a good remediation? Just switching to ArrayList? Freel free to submit PRs for possible improvements.

sebthom:
I am investigating that.

sebthom:
Please try the latest snapshot from https://download.eclipse.org/tm4e/snapshots/ it contains the performance improvements I merged.

sebthom:
[...]

PR #543, opened 2023-06-07
perf: faster processing of large file

sebthom:
This PR addresses #540 and additionally improves general rendering performance.

Full colorization of this file https://github.com/json-iterator/test-data/blob/master/large-file.json (26MB, 11k lines) took 29 seconds on my machine before this PR and takes 20 seconds with this PR applied. Also short UI freezes encountered without the PR are gone with the PR applied.


angelozerr:
@sebthom have you some benshmark about the performance improvement?

To be honnest with you I dont understand where you have optimized the code. Is it just use of ArrayList instead of LinkedArrayList,?

Many thankq for your clarification.

sebthom:
> @sebthom have you some benshmark about the performance improvement?
> 
> To be honnest with you I dont understand where you have optimized the code. Is it just use of ArrayList instead of LinkedArrayList,?
> 
> Many thankq for your clarification.

This [commit](https://github.com/eclipse/tm4e/pull/543/commits/02d4f0a7765ec983091b840dd522661a2f061ccc) replaces LinkedList with ArrayList in AbstractModelLines to address https://github.com/eclipse/tm4e/issues/540. AbstractModelLines mostly accesses internal list using random access via List#get(int). This results into UI freezes with a lot of lines.

There is actually only one use case where LinkedList is faster than ArrayList: when adding/removing elements at the beginning of a list. Even for insertions and removal within large lists ArrayList is faster because with LinkedList the list must be traversed up to the point where elements are removed/added.

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/AbstractModelLines.java
```diff
@@ -37,7 +37,7 @@ static final class ModelLine {
 		List<TMToken> tokens = Collections.emptyList();
 	}
 
-	private final List<ModelLine> list = new LinkedList<>();
+	private final List<ModelLine> list = new ArrayList<>();
```
szarnekow, line 40: This is a nice starting point.
I suggest to go a bit further and try to work based on addAll(int idx, Collection elements), and list.subList().clear();

Two examples:

```
protected void addLines(final int lineIndex, final int count) {
		if (count < 1)
			return;
		// prepare temporary list;
		var newLines =    new ArrayList<>(count)
		for (int i = 0; i < count; i++) {
			newLines.add(new ModelLine());
		}
		synchronized (list) {
			final var firstLine = getOrNull(lineIndex);
			list.addAll(lineIndex, newLines);
			if (firstLine != null) {
				list.get(lineIndex).startState = firstLine.startState;
			}
			updateLine(lineIndex);
		}
	}
```

similarly `removeLines` could benefit from `list.sublist(lineIndex, lineIndex + count).clear()`

szarnekow: Likewise there is potential for optimizations in replaceLines via `subList().replaceAll(..)` followed by an addAll or clear depending on the requested changes.

szarnekow:
> took 29 seconds on my machine before this PR and takes 20 seconds with this PR 

Could you please provide updated numbers? Because to be honest, 20 seconds is still quite slow.

org.eclipse.tm4e.core/src/main/java/org/eclipse/tm4e/core/model/AbstractModelLines.java
sebthom: > > took 29 seconds on my machine before this PR and takes 20 seconds with this PR
> 
> Could you please provide updated numbers? Because to be honest, 20 seconds is still quite slow.

I shaved of a few more seconds by adding a token caching on the hot path but I don't see more low hanging fruits that will have a noticeable. Because of it's async nature profiling is a bit difficult. Most of the time seems now to be spend in the joni library for TextMate regex searches and in eclipse UI classes doing the rendering. I see one more possible optimization but I am not sure how/if this will result in a noticeable performance improvement https://github.com/eclipse/tm4e/issues/38

Issue #561, opened 2023-06-16
Main thread locking overagressively


jonahgraham:
_Extracted from discussion in https://github.com/eclipse/tm4e/discussions/560_

When opening large files, like the example C++ file in #560 I see the main thread spending most of its time locked on an object, here is the stack trace of the two key threads:

```java
"main" #1 prio=6 os_prio=0 cpu=19432.94ms elapsed=82.09s tid=0x00007fd198028e90 nid=0x13a2e waiting for monitor entry  [0x00007fd1a00f4000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at org.eclipse.tm4e.core.model.AbstractModelLines.getOrNull(AbstractModelLines.java:147)
        - waiting to lock <0x0000000098580240> (a java.util.LinkedList)
        at org.eclipse.tm4e.core.model.TMModel.getLineTokens(TMModel.java:286)
        at org.eclipse.tm4e.ui.text.TMPresentationReconciler.colorize(TMPresentationReconciler.java:540)
        at org.eclipse.tm4e.ui.text.TMPresentationReconciler$InternalListener.colorize(TMPresentationReconciler.java:410)
        at org.eclipse.tm4e.ui.text.TMPresentationReconciler$InternalListener.lambda$0(TMPresentationReconciler.java:386)
        at org.eclipse.tm4e.ui.text.TMPresentationReconciler$InternalListener$$Lambda$1335/0x000000010096d970.run(Unknown Source)
        at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
        at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)

---

"org.eclipse.tm4e.ui.internal.model.TMDocumentModel" #103 daemon prio=1 os_prio=0 cpu=44439.56ms elapsed=47.43s tid=0x00007fd198795200 nid=0x13b48 runnable  [0x00007fd068fae000]
   java.lang.Thread.State: RUNNABLE
        at java.util.LinkedList.node(java.base@17.0.7/LinkedList.java:580)
        at java.util.LinkedList.get(java.base@17.0.7/LinkedList.java:481)
        at org.eclipse.tm4e.core.model.AbstractModelLines.getOrNull(AbstractModelLines.java:149)
        - locked <0x0000000098580240> (a java.util.LinkedList)
        at org.eclipse.tm4e.core.model.TMModel.getLineTokens(TMModel.java:286)
[...]

Issue #544, opened 2023-06-07
reformatting XML file removes lots of syntax highlighting


garretwilson:

sebthom:
Here is a minimal reproducer:
```xml
<foo>
   <foo>a</foo> <!-- VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY LONG COMMENT -->
   <bar>b</bar> <!-- VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY LONG COMMENT -->
</foo>
```

CTRL+ F results in:
![image](https://github.com/eclipse/tm4e/assets/426959/73f1e6b5-9a05-4671-8d92-e441698643a3)

sebthom:
This is a tricky race condition in the `TMModel` class between the main UI thread and the `TokenizerThread` on the `ModelLine.isInvalid` field. I checked the git history and I believe this issue exists since the beginning of the tm4e project. I'm trying to find a way to solve this without sacrificing performance. I'm thinking about ditching the `isInvalid` field altogether and only rely on the `TMModel.invalidLines` PriorityBlockingQueue.

angelozerr:
Please note the problem with syntax coloartion after a XML format occurs since XML language server provide a new XML formatter which remove / insert spaces instead of returning the full XML formatted content.

It means that when you format a XML format:

 * with old formatter gerenated just one LSP TextEdit
 * the new formatter generate several LSP TextEdit which causes this issue.

It should be nice to fix TM4E with several TextEdit changes in same time, but if it is too hard, we could add a new settings to XML LemMinX language server to return one TextEdit instead of several TextEdit. //cc @mickaelistria

sebthom:
I am trying to get it fixed for the new formatter. This XML formatter might not be the only one sending multiple text edits.

angelozerr:
Thats great, good luck @sebthom ! We should find a solution to write a test with this issue. If it is hard to reproduce it because of some async process perhaps we should disable those async process in a test ?

sebthom:
@garretwilson Please give the latest snapshot a try https://download.eclipse.org/tm4e/snapshots/

As explained in the related PR at https://github.com/eclipse/tm4e/pull/566#issue-1765085156 , there is still one minor issue where the styling is lost for some lines of reformatted multi-line xml comments (they are rendered with the default color) but at least the lines are not rendered with wrong colors. Anyways, the situation should be much improved now.

sebthom:
[...]

PR #566, opened 2023-06-20
fix: sync issues between TokenizerThread and TMPresentationReconciler

sebthom:
This PR is a major rewrite of how async line tokenization/colorizing is performed. It addresses the broken syntax highlighting #544 after reformatting XML files via the XML Language Server. It also further improves rendering performance and UI responsiveness issues as reported by #540 and #561.

---

There is one minor rendering issue left which I am unsure how to address. When multiline XML comments are reformatted, **some** of the lines that don't change due to reformatting loose their coloring. This is however not an issue caused by the tokenization model being out-of-sync or the TMPresentationReconciler not applying coloring to damaged regions but by the fact that these regions don't seem to be reported as damaged by the TextViewer. For performance reasons I am reluctant to simply re-colorize all lines of a document in case anything changed.

Here is an example:
image
Lines with a line number with white background did not change during the reformatting. Some of them loose their styling but not all.


sebthom:
@angelozerr do you have a chance to give this a try before I merge it?

angelozerr:
I'm sorry @sebthom it wil be hard for me to find time to test your PR.

It seems that you have add some colorize logic with DocumentListener, it means that colorisation will be done on backround and with document listener both?

sebthom:
> It seems that you have add some colorize logic with DocumentListener, it means that colorization will be done on backround and with document listener both?

I actually did not change this part. In the UI package I mainly did refactoring, like separating the colorization code into a dedicated separate class.

The main difference is that now when a lot of changes are performed in very fast succession (like done by the formatter), these changes are processed in a batch instead of separately and then firing a ModelTokensChangedEvent for each change. This confused the colorizer as it received line numbers that were outdated.

The other difference is that the managing of the TMModels internal list of lines is now completely down by the tokenizer thread whereas before the UI thread would effectively insert/remove lines which resulted in UI freezes for very large files. Now the UI thread only adds "edit" events to a non-blocking queue.

angelozerr:
Thanks so much for your clarification. It looks great! Congrats!

I wonder if we could add a test with TextEdit like we have for XML formatter?

sebthom:
@angelozerr I don't have much experience with Eclipse UI tests. Is it possible to start a UI test directly from within the Eclipse Workspace via `Run As -> JUnit Test`? So far I can only get the UI tests working when running them on the command line via maven which I find extremely cumbersome during development.

For example when I try to execute `TMinGenericEditorTest` or the `TMEditorColorTest` from within Eclipse all I get is `java.lang.NoClassDefFoundError: org/eclipse/e4/ui/workbench/IWorkbench`

Running the `RegistryTest` I get `java.lang.NullPointerException: Cannot invoke "org.eclipse.core.runtime.content.IContentTypeManager.getContentType(String)" because the return value of "org.eclipse.core.runtime.Platform.getContentTypeManager()" is null` and so on...

angelozerr:
See

image

angelozerr:
My idea was to write a test with MultiTextEdit like https://github.com/eclipse/lsp4e/blob/2a404e96ab462a1929a156b604745def685a166a/org.eclipse.lsp4e/src/org/eclipse/lsp4e/LSPEclipseUtils.java#L497

Issue #572, opened 2023-07-28
Theme that replicates the coloring of the WTP XML editor


kevloral:
Having a theme that replicated the coloring scheme used in the WTP XML editor would be nice.

angelozerr:
Any contribution are welcome!

kevloral:
The theme is called "Traditional" and it is quite close to the syntax coloring scheme used in the WTP XML editor. It is not perfect, though, as some elements that would colored differently in the original editor share the same CSS here. For the time being, the only difference I have been able to find can be seen on this screenshot, but there could be more:

![image](https://github.com/eclipse/tm4e/assets/3647248/25040129-576d-4b2f-b093-df66aa5a825e)

In the original WTP editor, the equals sign would be black, but here that sign shares the same CSS (.meta.tag) as the tag delimiters.

sebthom:
If the theme is specific for XML files, maybe it should be added to the language pack instead of the default UI plugin and be named "WTP XML Classic" or something like this?

kevloral:
You are both right. I named it _Traditional_ because, as you may imagine, I have started using this XML editor after having used the WTP XML editor for years. Therefore, I agree something along the lines of "WTP XML Classic" would be much more appropriate.

angelozerr:
> If the theme is specific for XML files,

It is specific for WTP, in other words CSS, XML, HTML.

> Therefore, I agree something along the lines of "WTP XML Classic" would be much more appropriate.

+1

> maybe it should be added to the language pack

I have not strong opinion about that

sebthom:
We could also just call the theme `WTP` if it was used by WTP for more than just XML files.

kevloral:
Please note that I have only tested the new theme with XML files.

sebthom:
@kevloral you seemed to have closed your PR. Is this enhancement not needed anymore?

kevloral:
My bad. I thought it had already been merged. I have a local copy of the .css file, so I will use it to create a new PR for this.

PR #573, opened 2023-07-28
feat: #572 Theme replicating the coloring of the WTP XML editor

kevloral:
It is not perfect (some items share the same CSS), but it is close enough.


angelozerr:
Thanks @kevloral !

Could you add a screenshot with a sample just to see the result of your PR please.

angelozerr:
Perhaps we should find a new name than Tradional? WTP for instance?

I suggest too to add your theme in the extensionpoint https://github.com/eclipse/tm4e/blob/88bb9c8d20b5c0bfb11573766fa709bba38323bf/org.eclipse.tm4e.ui/plugin.xml#L44 in order to TM4E provides this new theme.

kevloral:
The theme has been renamed (now it is called _WTP XML Classic_) and has been added to the plugin descriptor.

sebthom:
@kevloral can you please bump the minor version in the plugin.xml and the pom.xml of the tm4e.ui plugin. This should fix the build error

kevloral:
It seems Jenkins is failing because the artifact version has not been increased from the baseline (it is still 0.7.0). Would you like me to increase it (0.7.1?) or would you rather do it yourselves?

kevloral:
Sure, but I do not know the policy you have wrt versioning. Would 0.7.1 be fine? Or should it be 0.8.0?

kevloral:
Feature (_org.eclipse.tm4e.feature_) version updated (0.6.0 -> 0.7.0) too.

mickaelistria:
GitHub says "@kevloral closed this by deleting the head repository". Is this intended?

PR #576, opened 2023-08-04
build(deps): Build against new Orbit and Platform

akurtakov:
Mainly to pick up Guava 32.x to get rid of
https://nvd.nist.gov/vuln/detail/CVE-2023-2976

Version bumps as needed.


jcompagner:
which orbit platform do you build against then? there is already a new one that has this guava?

because the current snapshot break now: https://github.com/eclipse/tm4e/discussions/580

i guess if the snapshot repo did also provide those jars it could work, but for now i think for most people the snapshot dir is unusable.

akurtakov:
@jcompagner Orbit changed significantly for 2023-09 thus I use https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2023-09 now . See https://www.eclipse.org/lists/cross-project-issues-dev/msg19699.html for details.

PR #599, opened 2023-10-18
build: Build against Eclipse 4.29

akurtakov:


.github/workflows/licensecheck.yml
```diff
@@ -1,113 +1,22 @@
...
+name: License vetting status check
 
 on:
   push:
-    branches-ignore: # build all branches except:
-      - 'dependabot/**' # prevent GHA triggered twice (once for commit to the branch and once for opening/syncing the PR)
-    tags-ignore: # don't build tags
-    - '**'
+    branches: 
+      - 'main'
```
sebthom, line 8: I don't think the license check should only run on the main branch. PRs should only be merged once the license check was successful.

Issue #661, opened 2024-01-11
Lot of errors logged on non existing resources since 0.9.0


iloveeclipse:
I've tried to update from 0.8.1 to 0.9.0 but in our automates tests we see huge number of fails because the code now logs lot of errors during testing.

The problem is that the content type checks that run in UI thread are running asynchronously to the resource changes and if tests create/delete/move files the `HasLanguageConfigurationPropertyTester` may be queried on stale resources, flooding the log with errors.

```
!ENTRY org.eclipse.tm4e.ui 4 0 2024-01-11 00:04:40.410
!MESSAGE Resource '/xxx.txt' does not exist.
!STACK 1
org.eclipse.core.internal.resources.ResourceException(/xxx.txt)[368]: java.lang.Exception: Resource '/xxx.txt' does not exist.
	at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(ResourceException.java:42)
	at org.eclipse.core.internal.resources.ResourceException.<init>(ResourceException.java:38)
	at org.eclipse.core.internal.resources.Resource.checkExists(Resource.java:367)
	at org.eclipse.core.internal.resources.Resource.checkAccessible(Resource.java:241)
	at org.eclipse.core.internal.resources.File.getContentDescription(File.java:276)
	at org.eclipse.core.internal.filebuffers.ResourceTextFileBuffer.getContentType(ResourceTextFileBuffer.java:187)
	at org.eclipse.tm4e.ui.internal.utils.ContentTypeHelper.getContentTypes(ContentTypeHelper.java:109)
	at org.eclipse.tm4e.ui.internal.utils.ContentTypeHelper.findContentTypesFromFileBuffers(ContentTypeHelper.java:92)
	at org.eclipse.tm4e.ui.internal.utils.ContentTypeHelper.findContentTypes(ContentTypeHelper.java:55)
	at org.eclipse.tm4e.languageconfiguration.internal.HasLanguageConfigurationPropertyTester.test(HasLanguageConfigurationPropertyTester.java:46)
	at org.eclipse.core.internal.expressions.Property.test(Property.java:65)
	at org.eclipse.core.expressions.TestExpression.evaluate(TestExpression.java:107)
	at org.eclipse.core.expressions.CompositeExpression.evaluateAnd(CompositeExpression.java:54)
	at org.eclipse.core.expressions.WithExpression.evaluate(WithExpression.java:84)
[...]

PR #663, opened 2024-01-11
fix: many Exceptions logged if editor underlying resource (re)moved #661

sebthom:


org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/utils/ContentTypeHelper.java
```diff
@@ -131,7 +131,10 @@ private static ContentTypeInfo getContentTypes(final ITextFileBuffer buffer) {
 			} catch (final Exception e) {
 				return null;
 			}
-		} catch (final CoreException | IOException ex) {
+		} catch(FileNotFoundException ex) {
```
iloveeclipse, line 134: This will not help. The error is not FileNotFoundException but CoreException, so it willl be still logged below

sebthom: OK, I changed it to a catch all but log the exception when tracing is enabled.

PR #690, opened 2024-02-09
fix: return correct value of "dark" property for default Eclipse theme

joaodinissf:
On a new workspace, prefs.get(PreferenceConstants.E4_THEME_ID, null) will return null, because no theme changes have yet taken place.

In this case (i.e., when the default theme is active), it is sensible to default to the system's dark mode setting, instead of defaulting to the light mode.

### DIFF
--- a/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/utils/PreferenceUtils.java
+++ b/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/utils/PreferenceUtils.java
@@ -11,12 +11,16 @@
  */
 package org.eclipse.tm4e.ui.internal.utils;
 
+import static org.eclipse.ui.themes.IThemeManager.DEFAULT_THEME;
+
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.swt.widgets.Display;
 import org.eclipse.tm4e.ui.TMUIPlugin;
 import org.eclipse.tm4e.ui.internal.preferences.PreferenceConstants;
+import org.eclipse.ui.PlatformUI;
 
 public final class PreferenceUtils {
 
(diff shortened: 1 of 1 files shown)


org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/utils/PreferenceUtils.java
```diff
@@ -32,7 +36,14 @@ public final class PreferenceUtils {
...
 		final IEclipsePreferences prefs = getE4PreferenceStore();
 		final String eclipseThemeId = prefs == null ? null : prefs.get(PreferenceConstants.E4_THEME_ID, null);
-		return eclipseThemeId != null && eclipseThemeId.toLowerCase().contains("dark");
+
+		if (eclipseThemeId != null) {
+			return eclipseThemeId.toLowerCase().contains("dark");
+		}
+
+		// If the default Eclipse theme is active, return the system's dark mode setting
+		return DEFAULT_THEME == PlatformUI.getWorkbench().getThemeManager().getCurrentTheme().getId() //
```
mickaelistria, line 45: Wouldn't it be more reliable to query something like `PlatformUI.getActiveWorkbenchWindow().getShell().getBackground()` here to establish whether the theme is dark or not?

sebthom: I think this is a good idea, we could use something like here to decide if it counts as a dark color:

https://github.com/eclipse/tm4e/blob/945f958d43469c1be511e90d63c91cf6787f13b3/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/AbstractThemeManager.java#L114-L118

sebthom:
Dear João, first thanks for this PR, we really appreciate any contributions. While looking into the issue I found another problem with color themes and created PR #693 for this. This PR should also solve the issue you are facing by employing a background color brightness check. Please try the latest snapshot version and check if it now works for you.

joaodinissf:
Your PR looks good to me--thanks for the quick feedback/fix!

The issue that led me to create this PR has indeed been fixed, so I am closing it now.

PR #714, opened 2024-02-16
ci: test against minimum supported and current stable Eclipse release

sebthom:
This PR changes the github action workflow in a way that TM4E builds are tested against three different platform targets:
- the oldest supported Eclipse version (currently 4.29 aka 2023-09)
- the latest/current stable Eclipse version (4.30 aka 2023-12)
- the newest integration build of the next Eclipse version (4.31 aka 2024-03)

I'd like to try maintain compatibility with one or two Eclipse versions back as long as no security concerns or functional changes in the Eclipse platform require an immediate upgrade to the latest Eclipse platform. These new matrix builds help with that.


akurtakov (approved):
Change looks good as long as there is someone to work on keeping such compatibility working. I personally can not care for anything older than latest stable release.

Issue #713, opened 2024-02-16
Default theme can't be specified via plugin customization


iloveeclipse:
The code below only asks for a **user** preference, not for **all** possible configured preferences like plugin customization.
https://github.com/eclipse/tm4e/blob/e87d06b6a2afd5fbb2b68a91345508b1ce09b2ac/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/internal/themes/ThemeManager.java#L125

So plugin customization file that contains something like this doesn't work:
```
# Use "classic" theme for textmate
org.eclipse.tm4e.ui/org.eclipse.tm4e.ui.themes.defaultLightTheme=org.eclipse.tm4e.ui.themes.WtpXmlClassic
```

The right way is to read preferences from preferences service and write to the instance scope.

sebthom:
Before I start researching into Eclipse preferences. Could you give me a few lines of pseudo code, then I can change this.

iloveeclipse:
```
IPreferencesService preferenceService = Platform.getPreferencesService();
String value = preferenceService.getString(PLUGIN_ID, KEY, DEFAULT_VALUE, /* search in all scopes is null */ null);
```

iloveeclipse:
Verified fix https://github.com/eclipse/tm4e/pull/716 : works fine in 0.10.1.202402162114 version & product customization.

iloveeclipse:
@sebthom : do you plan to release 0.10.1 may be? 
I'm particularly interested in https://github.com/eclipse/tm4e/issues/709 & https://github.com/eclipse/tm4e/issues/713 fixes.

sebthom:
> @sebthom : do you plan to release 0.10.1 may be? I'm particularly interested in #709 & #713 fixes.

I just released it

PR #716, opened 2024-02-16
fix: default theme cannot be set in plugin customization file #713

sebthom:


sebthom:
@iloveeclipse WDYT?

iloveeclipse:
Have no IDE at hand, but looks exactly what I would expect.

PR #726, opened 2024-03-05
fix: BadLocationExceptions in Colorizer event processing.

andrewL-avlq:
The Colorizer should only process events for the document currently present in the viewer. This prevents BadLocationExceptions occurring when the content of the viewer changes rapidly.


sebthom:
Nice, thanks!!

PR #752, opened 2024-05-16
feat: add option to display TextMate token info in hover

sebthom:
This PR adds the ability to show the type and scopes of TextMate tokens in a hover which is useful when developing TM themes and needing to style specific elements.

image
image


angelozerr:
It is a fantastic feature! Please note that vscode provides this features too https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#textmate-tokens-and-scopes andit display another relevant information 

image

sebthom:
@angelozerr We currently don't have all this information available. However, I added a second commit that extends the hover like this:

image

angelozerr:
Thanks @sebthom for your improvement.

I suggest that you write a doc about this great feature

PR #826, opened 2024-11-04
ci: add codeql workflow

sebthom:


netomi:
should work now.

netomi:
The Code Scanning Tab looks good now. I had to manually delete the old failing configurations from the default setup. Its so intuitive how they designed that and how it works, but we are all green now.

PR #828, opened 2024-11-06
fix: Incomplete URL scheme check in marked.js

sebthom:
Fixes [https://github.com/eclipse-tm4e/tm4e/security/code-scanning/2](https://github.com/eclipse-tm4e/tm4e/security/code-scanning/2)

To fix the problem, we need to extend the URL scheme check to include `data:` in addition to `javascript:` and `vbscript:`. This will ensure that URLs with potentially harmful schemes are properly sanitized. The best way to fix this without changing existing functionality is to modify the condition on line 878 to also check for `data:` schemes.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/org.eclipse.tm4e.markdown/marked.js/marked.js
+++ b/org.eclipse.tm4e.markdown/marked.js/marked.js
@@ -875,7 +875,7 @@ Renderer.prototype.link = function(href, title, text) {
     } catch (e) {
       return '';
     }
-    if (prot.indexOf('javascript:') === 0 || prot.indexOf('vbscript:') === 0) {
+    if (prot.indexOf('javascript:') === 0 || prot.indexOf('vbscript:') === 0 || prot.indexOf('data:') === 0) {
       return '';
     }
   }


angelozerr:
You can remove this marked.js

I did that a long time ago to play with marked.js with the marked.html file.

Indeed I had ported the markdown.js to Java to try to have the same renderer than vscode.

angelozerr:
Just to give you some info about this tm4e markdown plugin.

My goal was to have the same renderer than vscode to support hover (in LSP4E for instance). It is the reason why I have translated marked.js in Java because it is used in vscode.

And the nice thing is that markdown code block is managed with TM4E tokenizer, it means that it uses textmate to manage syntax coloration in hover.

I had used this marked Java port in my very old project at https://github.com/angelozerr/typescript.java/blob/8a6759ca825c2b0ae14cb8e09e5ba1778aae24d3/eclipse/ts.eclipse.ide.ui/src/ts/eclipse/ide/ui/utils/HTMLTypeScriptPrinter.java#L86

sebthom:
This relates to https://github.com/eclipse-tm4e/tm4e/issues/286 right?

angelozerr:
> This relates to https://github.com/eclipse-tm4e/tm4e/issues/286 right?

Yes:) I have forgotten that we have already a discussion about that

PR #855, opened 2025-01-07
fix: FoldingRules.toString to print markers

akurtakov:
Markers label was added but the actual value wasn't.


akurtakov:
Merging as this is plain bugfix in non critical code.

Issue #408, opened 2022-06-03
Argument not valid error when shifting JSON left


PyvesB:
Follow up to #236.

Consider the following partial JSON snippet (note the spacing to the left):
```
    [
        { "key1": "value1",
        "keys2" }
    ]
```

Paste it in an empty file, select all of it, and press the Shift + Tab keys to push the JSON to the left. The following stack trace appears in the error log:
```
java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4899)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.custom.StyledText.setStyleRanges(StyledText.java:10651)
	at org.eclipse.swt.custom.StyledText.replaceStyleRanges(StyledText.java:8239)
	at org.eclipse.jface.text.TextViewer.addPresentation(TextViewer.java:4773)
	at org.eclipse.jface.text.TextViewer.changeTextPresentation(TextViewer.java:4850)
	at org.eclipse.tm4e.ui.text.TMPresentationReconciler.applyTextRegionCollection(TMPresentationReconciler.java:688)
	at org.eclipse.tm4e.ui.text.TMPresentationReconciler.colorize(TMPresentationReconciler.java:592)
	at org.eclipse.tm4e.ui.text.TMPresentationReconciler$InternalListener.textChanged(TMPresentationReconciler.java:335)
	at org.eclipse.jface.text.TextViewer.updateTextListeners(TextViewer.java:2785)
	at org.eclipse.jface.text.TextViewer.fireRedrawChanged(TextViewer.java:4995)
	at org.eclipse.jface.text.TextViewer.enabledRedrawing(TextViewer.java:5035)
	at org.eclipse.jface.text.TextViewer.enabledRedrawing(TextViewer.java:5003)
	at org.eclipse.jface.text.TextViewer.setRedraw(TextViewer.java:5085)
	at org.eclipse.jface.text.TextViewer.setRedraw(TextViewer.java:5059)
	at org.eclipse.jface.text.TextViewer$RewriteTarget.setRedraw(TextViewer.java:1004)
	at org.eclipse.jface.text.TextViewer$DocumentRewriteSessionListener.documentRewriteSessionChanged(TextViewer.java:1410)
	at org.eclipse.jface.text.AbstractDocument.fireRewriteSessionChanged(AbstractDocument.java:1432)
[...]

PR #856, opened 2025-01-10
fix: IllegalArgumentException in Colorizer event processing.

andrewL-avlq:
The Colorizer should not process stale tokens if they go over the end of the line in the current document. This prevents IllegalArgumentException occurring when the Tokenizer thread runs slow or is delayed relative to the textChanged notification in the reconciler.

### DIFF
--- a/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/text/Colorizer.java
+++ b/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/text/Colorizer.java
@@ -166,6 +166,13 @@ class Colorizer {
 			length = Math.min(length, damageRegion.getOffset() + damageRegion.getLength() - lastStart);
 			addStyleRange(presentation, lastStart, length, lastAttribute);
 			viewer.changeTextPresentation(presentation, false);
+		} catch (final IllegalArgumentException | BadLocationException ex) {
+			error = ex;
+			// These exceptions can be thrown if there is a delay running the tokenizer thread
+			// and the tokens become out of sync with the document line data.
+			// As this is an expected state, only log them if tracing is enabled.
+			if (TMUIPlugin.isLogTraceEnabled())
+				TMUIPlugin.logError(ex);
 		} catch (final Exception ex) {
 			error = ex;
 			TMUIPlugin.logError(ex);


andrewL-avlq:
I think this should fix issues like #408.

sebthom:
Thanks for your PR. However I am not sure if this is the right fix or if we should better suppress the IllegalArgumentException as well as the BadLocationExceptions in the catch block and only log them when tracing is enabled. This new check does not fully prevent the occurrence of race conditions in the following lines. Also I am not sure if continuing to colorize subsequent lines makes sense if the tokens of one line are stale.

andrewL-avlq:
Sure, this PR is a starting point for discussion if nothing else.

The edits that the Tokenizer is processing might not include changes to the following lines, so exiting the colorizer early didn't seem quite right. On the other hand, when the Tokenizer thread has updated the tokens, the colorizer will run again, so I am fine with throwing.

Would you rather catch the IllegalArgumentException, or detect the condition early (such as having getTokenLength detect that it is about to return a negative value) and throw a custom exception (maybe even BadLocationException)?

Given that the Tokenizer thread runs asynchronously, I was not trying to prevent any racing, but rather to have the colorizer cope with what goes wrong if the model is out of sync with the document (as there already appear to be conditions to handle other cases). Also for this reason, I agree that the logging of the BadLocationExceptions should be guarded and only logged when tracing is enabled.

sebthom:
I think I would just add another `} catch (final IllegalArgumentException | BadLocationException ex) { error = ex; TMUIPlugin.logTrace(ex); }` block with some explanation before 

https://github.com/eclipse-tm4e/tm4e/blob/83e998cb976e4762b31d1ed285c2d6a8830582c8/org.eclipse.tm4e.ui/src/main/java/org/eclipse/tm4e/ui/text/Colorizer.java#L169-L172

sebthom:
Thanks!

andrewL-avlq:
No problem.
What is the timescale for seeing this in a release?

sebthom:
There is no fixed plan. I can create a release when the license check works again. I already opened a support ticket.

andrewL-avlq:
That would be great thanks.

sebthom:
@andrewL-avlq A new release 0.14.1 is available

Issue #827, opened 2024-11-05
TextMate syntax coloring doesn't appear on first toggle split editor


sdaimwood:
Our organization maintains several language servers that use TM4E in an eclipse-based product. We've had customer reports and verified that when first using the `Toggle Split Editor` view, the newly opened window does not highlight until it is toggled on a second time. 

Using the publicly available YAML editor as an example: 

1) Open YAML source in the YAML Editor with Window/Editor/Toggle
Split Editor (Vertical).
The new second window will have no syntax highlighting.

2) Without closing, use same YAML source with Window/Editor/Toggle Split Editor
(Horizontal).
The new horizontally split window does have syntax highlighting.

3) Use same YAML source with Window/Editor/Toggle Split Editor
(Vertical).
The new vertically split screen will have syntax highlighting.

It appears that for each file, only the first invocation of split screen editing doesn't have highlighting, but every following invocation does. 

Environment: 
- Eclipse 4.31
- TM4E 0.13.0, LSP4E 0.18.4

sebthom:
I can reproduce this with the latest eclipse and tm4e release.

@mickaelistria any idea how to address this?

sebthom:
I cannot reproduce this anymore with Eclipse 2024-12 (4.34.0) and TM4E 0.14.1.

Feel free to reopen if issue persists.

sdaimwood:
Thanks for the update, unfortunately I'm still seeing this running Eclipse 4.34 and TM4E 0.14.1, was running 0.14.0 and even updated to double check.

sdaimwood:
@sebthom I am on macOS if that makes a difference, but have had colleagues test on Windows as well and still seeing the same issue.

sebthom:
I downloaded/extracted/run vanilla eclipse-java-2024-12-R-win32-x86_64.zip on Windows 10.
The error is not reproducible for me. So even if we reopen I don't see this to be actionable.

![Image](https://github.com/user-attachments/assets/c32dd17f-7113-47c5-a4dd-0e36490b2d4a)

sdaimwood:
Very odd. Here's what I'm seeing. 

![Image](https://github.com/user-attachments/assets/296f39ca-8595-45a4-b623-9ac83ca83482)

Just tested on a Windows 10 VM after a fresh install of vanilla 2024-12 and seeing the issue there still as well. Do you have any idea what I might be missing in an environment config?

sebthom:
[...]

PR #870, opened 2025-03-02
fix: Invoke colorizer when editor is cloned or splitted #827

oleosterhagen:
For new models the colorizer will be invoked after tokenization. For reused models (e.g. when splitting an editor) this must be done explicitly.


sebthom (approved):

sebthom:
Thanks!

Issue #147, opened 2017-11-05
Usage of term "snippet"


amyers735:
I just wanted to raise a comment about the usage of the term "snippet" for the language preview in the prefs.  

Given that TextMate has a concept of snippets (see https://manual.macromates.com/en/snippets) I think this could be a bit confusing.  Is the idea that the snippets in tm4e could later be used the same was as TextMate snippets?  Or are they only ever going to be for syntax highlight previews?

angelozerr:
> Given that TextMate has a concept of snippets (see https://manual.macromates.com/en/snippets) I

I didn't know that.

> I think this could be a bit confusing. 

yes you are right.

> Is the idea that the snippets in tm4e could later be used the same was as TextMate snippets? Or are they only ever going to be for syntax highlight previews?

It's just for `syntax highlight previews`.

@mickaelistria I think we should rename that like `previewSnippet`, no?

@am2605 what is your suggestion?

am2605:
previewSnippet sounds okay to me.  Or codeSample perhaps?.  I'm not sure. :)

mickaelistria:
Whichever would work for me. @am2605 Would you please contribute a PR?

sebthom:
Addressed by #892

PR #892, opened 2025-05-31
refact: deprecate snippets ext. point and add samples ext. point

sebthom:
This addresses #147. The old "snippets" extension point is still functional so no immediate change required by grammar providers.

Old:
```xml
   <extension
         point="org.eclipse.tm4e.ui.snippet">
      <sample
            name="org.eclipse.tm4e.ui.tests.snippet1"
            path="grammars/snippet.ts"
            scopeName="source.ts">
      </sample>
   </extension>
```
New:
```xml
   <extension
         point="org.eclipse.tm4e.ui.samples">
      <sample
            name="org.eclipse.tm4e.ui.tests.sample1"
            path="grammars/sample.ts"
            scopeName="source.ts">
      </sample>
   </extension>
```


mickaelistria:
I'm fine with such renames. Theorically, we'd also keep ISnippet and ISnippetManager are deprecated; but if we can ensure enough those have no users, it's OK to delete them too.

sebthom:
> I'm fine with such renames. Theorically, we'd also keep ISnippet and ISnippetManager are deprecated; but if we can ensure enough those have no users, it's OK to delete them too.

I already did a thorough search on Google and GitHub and could not find any references to ISnippet and ISnippetManager in any external code so I think it is safe to rename them.

mickaelistria:
OK, good.

PR #897, opened 2025-06-02
feat: add folding reconciler

sebthom:


sebthom:
That indeed is an interesting question. I gave it a try, the TMFoldingStrategy applied the language-configuration.json foldings and the CSS language server applied it's own overlapping foldings. they get merged. even if both apply foldings for the same regions you only see one folding indicator.

sebthom:
@angelozerr I updated the code so that the folding reconciler is deactivated when an LS is available for a document.

mickaelistria:
> I updated the code so that the folding reconciler is deactivated when an LS is available for a document.

Could it be instead deactivated if some other folding annotations, not coming from TM4E, are present?

sebthom:
> Could it be instead deactivated if some other folding annotations, not coming from TM4E, are present?

@mickaelistria This is not possible I believe. The order in which reconcilers are applied is not predictable. I guess depending on plugin load order.

The TM4E folding reconciler activation now works as follows:
If folding markers are found in a `language-configuration.js` file or the TextMate grammar - and **no** language server is associated with the TextViewer - TM4E applies custom folding based on whitespace indentation and the configured fold markers.

Issue #949, opened 2025-11-03
Cannot indent on a newline when "Insert spaces for tabs" preference is active


ddscharfe:
## Bug Description
There seems to be a bug in which the indentation of a line is deleted when spaces are used instead of tabs.
The user cannot add indentation using the TAB key when the cursor is placed at the beginning of a blank line. If the cursor is placed at the end of a line that contains 4 spaces, the indentation is deleted.

## How to reproduce
1. Enable General->Editors->Text Editors->"Insert spaces for tabs" preference <img width="1005" height="512" alt="Image" src="https://github.com/user-attachments/assets/6c1deb79-54d0-4e01-85de-aca3a6264c6a" />
2. Create a project with an example file (e.g. a C++ file; it is important to use a language that has indentation rules in its language-configuration.json)
3. Place the cursor inside an indented line and press the TAB key
<img width="531" height="161" alt="Image" src="https://github.com/user-attachments/assets/974c23b1-2a85-4e33-b5fe-bfa3162d1262" />
<img width="531" height="125" alt="Image" src="https://github.com/user-attachments/assets/6dcbf42d-51e5-4cb1-b59f-9c57290e0273" />

### Analysis
Related Classes: 
- `org.eclipse.tm4e.languageconfiguration.internal.utils.TextUtils`
- `org.eclipse.tm4e.languageconfiguration.internal.LanguageConfigurationAutoEditStrategy`

The problem occurs for C++ since the last language pack update (commit 6bccc035e4d858dffcd7d7a356ea486c69a8739c, `/org.eclipse.tm4e.language_pack/syntaxes/cpp/cpp.language-configuration.json`). This update added indentation rules.

`LanguageConfigurationAutoEditStrategy::customizeDocumentCommand` handles the TAB key similarly to a paste operation and when there are indentation rules, it calls `TextUtils.replaceIndent`, which returns an empty `CharSequence`, effectively removing the existing spaces.

PR #951, opened 2025-11-03
fix: keep Tab-as-spaces indentation on blank lines (LangCfg)

sebthom:
Pressing Tab with "Insert spaces for tabs"inserted spaces that were treated like a paste and reindented via `replaceIndent(..., false)`, which returned an empty string and removed the indentation.

Skip reindentation for single-line, whitespace-only inserts so user-inserted spaces are preserved on blank lines. Multi-line pastes and non-whitespaceinserts remain unchanged.


sebthom:
@ddscharfe This PR is to address the issue you reported in #949 can you please try to install this snapshot version of TM4E and confirm that the issue is solved: https://github.com/eclipse-tm4e/tm4e/actions/runs/19042992902/artifacts/4452012289

PR #1004, opened 2026-04-10
feat: support multiline/block comments for C/C++

### LINKED ISSUE #1003: Support Multiline/Block Comments for C/C++
In CDT-LSP we have this [issue](https://github.com/eclipse-cdt/cdt-lsp/issues/507) to support multiline comments like:
```C
/*
 * 
 */
```
Since TM4E handles On Enter rules, I assume it would be the best to extend the `c.language-configuration.json` and `cpp.language-configuration.json`

ghentschke:
Fixes #1003

Extending the onEnterRules

### DIFF
--- a/org.eclipse.tm4e.language_pack/syntaxes/cpp/c.language-configuration.json
+++ b/org.eclipse.tm4e.language_pack/syntaxes/cpp/c.language-configuration.json
@@ -122,11 +122,38 @@
 			},
 			"afterText": {
 				"pattern": "^(?!\\s*$).+"
+				
 			},
 			"action": {
 				"indent": "none",
 				"appendText": "// "
 			}
 		},
+		// Start of /* or /** -> create block comment and indent
+		{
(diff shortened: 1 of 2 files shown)


ghentschke:
I am not sure if and which unit tests should be updated

sebthom:
The syntax and language configuration files are sourced directly from vscode repo. We do not modify upstream files in tm4e. If I would merge this, the changes would be gone the next time we sync with vscode. So this change should be better made in the source repo. Either at https://github.com/microsoft/vscode/tree/main/extensions/cpp or even in https://github.com/jeff-hykin/better-cpp-syntax

ghentschke:
Thanks for the feedback. It seems that the configuration comes from vscode. Haven't found any language-configuration.json or something similar in the jeff-hykin repo.

ghentschke:
Closed as it has to be fixed in vscode

PR #1018, opened 2026-07-04
build: Add org.eclipse.orbit.xml-apis-ext to category.xml

merks:
- Remove explicit *.source from *.target because includeSource="true" implies that.
- Update staging.target to work with latest 4.41 I-bulds.

https://github.com/eclipse-simrel/simrel.build/issues/1432

This ensures that the TM4E update site's dependencies are transitively more complex.

### DIFF
--- a/org.eclipse.tm4e.repository/category.xml
+++ b/org.eclipse.tm4e.repository/category.xml
@@ -12,6 +12,15 @@
    <bundle id="com.google.gson" version="0.0.0">
       <category name="Dependencies"/>
    </bundle>
+   <!--
+     This is a transitive dependency of org.apache.batik.css.
+     It provides package org.w3c.dom.svg version 1.1.0 and org.w3c.css.sac version 1.3.0.
+     Without this bundle in the update site, tm4e cannot be installed unless those packages are provided by some other update site.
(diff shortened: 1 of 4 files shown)


org.eclipse.tm4e.repository/category.xml
```diff
@@ -12,6 +12,9 @@
    <bundle id="com.google.gson" version="0.0.0">
       <category name="Dependencies"/>
    </bundle>
+   <bundle id="org.eclipse.orbit.xml-apis-ext" version="0.0.0">
```
sebthom, line 21: From you PR description I don't understand why we need this. Could you please elaborate and add an inline comment in this xml file so the entry does not get removed accidentally in future because AFAIK we do not reference this anywhere in our feature or plugin config files.

merks: Comments added.  

You're probably not aware from from where your package imports originate.  These are the dependencies on that bundle:

image

Note only do you have transitive dependencies on packages provided by this bundle, you have direct dependencies on those packages too.

image

You didn't need to be aware because the Platform's CSS engine always installed these things, but that's no longer true, so now you need to be aware.

merks:
Note that resolving the comment kind of hides all my comments....

sebthom:
Thanks!

PR #1059, opened 2026-09-21
ci: group codeql-action updates into a single PR

akurtakov:
Dependabot treats github/codeql-action/init and github/codeql-action/analyze as separate dependencies and opens one PR for each. Since each PR only bumps one of the two pins in codeql.yml, the CodeQL workflow fails with an init/analyze version mismatch until both are merged.

Group all github/codeql-action* updates so they are bumped together.

Assisted-by: Anthropic Claude Code (claude-opus-5[1m])


akurtakov:
Merge as it will have effect only after hitting main.

PR #1061, opened 2026-09-21
feat: add workspace language defaults and per-file choices

sebthom:
Support importing grammars and binding them to content types. Refresh highlighting and editing rules when a file's selected language changes.


org.eclipse.tm4e.registry/src/main/java/org/eclipse/tm4e/registry/internal/AbstractGrammarRegistryManager.java
```diff
@@ -307,9 +380,16 @@ void registerInjection(final TMScope scopeName, final String injectTo) {
 	@Override
 	public @Nullable Collection<IContentType> getContentTypesForScope(final ITMScope scope) {
 		// -> used by GrammarPreferencePage.createGrammarListContent().fillContentTypeTab()
-		return contentTypeToScopeBindings.values().stream()
+		final var contributed = contentTypeToScopeBindings.values().stream()
```
jensli, line 382: I myself think that code gets much easier to read if explicit types are used instead of `var` in places like this.

sebthom:
@jensli did you give it a try? is it what you expected functionality wise?

jensli:
> did you give it a try? is it what you expected functionality wise?

Yes, it works! Great!

Here are some thought and observations:

* I think it would be useful to define a command for the *Choose Language....* menu action. In that way it could be accessed using the *Quick Access* function (Ctrl + 3), or the user could set their own keyboard shortcut.

* The content type work is unrelated to the *Choose Language* functionality, isn't it? *Choose Language* is for choosing any language for a given file extension. The content type association is for associating languages and file extensions in a new way, but doesn't help with choosing languages. Or have I misunderstood something?

* It seems like this change also repaired the *Import TextMate Grammar* functionality. Before this change I got an exception  when trying to import.

* After this change, the new custom Mumps language was chosen by default, instead of the Objective-C language from the language pack, without having to manually switch language. I think that's a good thing. If the user adds a custom language that's probably the one they want to use.

sebthom:
> Here are some thought and observations:
> 
> * I think it would be useful to define a command for the _Choose Language...._ menu action. In that way it could be accessed using the _Quick Access_ function (Ctrl + 3), or the user could set their own keyboard shortcut.

That's a good idea, thanks!

> * The content type work is unrelated to the _Choose Language_ functionality, isn't it? _Choose Language_ is for choosing any language for a given file extension. The content type association is for associating languages and file extensions in a new way, but doesn't help with choosing languages. Or have I misunderstood something?

They serve different purposes, but are related. `Choose Language...` changes the language for one file, while content-type bindings set the default for matching files across the workspace.

Those associations also help put likely matches first in the chooser, alongside the grammar's own file patterns. And they let editing features such as bracket matching and folding follow the selected language.

> * It seems like this change also repaired the _Import TextMate Grammar_ functionality. Before this change I got an exception  when trying to import.

The import repair was a separate fix in https://github.com/eclipse-tm4e/tm4e/pull/1053

> * After this change, the new custom Mumps language was chosen by default, instead of the Objective-C language from the language pack, without having to manually switch language. I think that's a good thing. If the user adds a custom language that's probably the one they want to use.

This should actually only happen if you select the checkbox "Set this grammar as the workspace default for matching files" which is off by default.

sebthom:
@jensli please give the new snapshot version in https://github.com/eclipse-tm4e/tm4e/actions/runs/35843669695/artifacts/10742691263 a try. it contains the command handler change.

